PR #15, opened 2018-11-15
Update integration tests

guybedford:
This gets the integration tests running similarly to the webpack branch, so that we can bring these through easily.


guybedford:
I've updated this PR to properly display the error traces against the original sources with full source maps support.

guybedford:
I've updated this PR to use jest as with the webpack branch.

PR #20, opened 2018-11-16
Mark modules that could not be resolved as external

### LINKED ISSUE #16: Conditional requires break modules (webpack)
Affecting:
- `pg`
- `redis`

timneutkens:
Fixes #16

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,7 @@ module.exports = async (entry, { minify = true } = {}) => {
     externals: (context, request, callback) => {
       resolve(request, { basedir: context, preserveSymlinks: true }, (err) => {
         if (err) {
-          console.error(`ERROR RESOLVING MODULE ${request}, ADDING AS EXTERNAL`, err)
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -17,6 +17,16 @@ module.exports = async (entry, { minify = true } = {}) => {
       filename: "out.js",
       libraryTarget: "commonjs2"
     },
+    externals: (context, request, callback) => {
+      resolve(request, { basedir: context, preserveSymlinks: true }, (err) => {
```
guybedford: Another way to do this might be to put this into a `resolveId` plugin hook (a local inline plugin). `null` returns from resolveId are treated as externals. In the plugin hook, you'd then have access to the global resolver through `this.resolveId` so that it exactly matches the internal resolve implementation (no need to worry about divergence between `require('resolve')` and rollup resolve.

timneutkens: This is using webpack though 🕵️

guybedford: Ah sorry, missed that!

timneutkens: I assume something similar can be done using webpack, but this was the least time-consuming way to make it work 👍

PR #23, opened 2018-11-19
Add explicit file extensions to resolve

### LINKED ISSUE #22: Require is unable to resolve JSON files without explicit ".json" extension
For example, I'm trying to build an app with FIrebase Admin SDK, and it fails.

`npx ncc index.js -o bundle.js`
```
Module directory "/node_modules/@google-cloud/firestore/build/src/v1beta1" attempted to require "./firestore_client_config" but could not be resolved, assuming external.
```

`/node_modules/@google-cloud/firestore/build/src/v1beta1` folder looks like this:
screenshot 2018-11-19 at 11 44 59

And the `require()` looks like this `const gapicConfig = require('./firestore_client_config');`

nik-garmash:
Fixes #22 
Fixes https://github.com/zeit/now-builders/issues/52


index.js
```diff
@@ -3,6 +3,23 @@ const fs = require("fs");
 const webpack = require("./webpack/lib/webpack");
 const MemoryFS = require("memory-fs");
 
+function resolveModule (context, request, callback) {
+  const resolveOptions = {
+    basedir: context,
+    preserveSymlinks: true,
+    extensions: ['.js', '.json']
```
timneutkens: I think we need to add all file extensions webpack uses

rauchg:
@nik-garmash you are amazing. Thank you!

index.js
timneutkens: https://webpack.js.org/configuration/resolve/#resolve-extensions

```
extensions: ['.wasm', '.mjs', '.js', '.json']
```

rauchg:
By the way: is this enumeration in sync with webpack's? I believe webpack supports other extensions. e.g.: `.mjs`

rauchg:
@timneutkens I can take care of adding a few tests. We're seeing it with `@google/datastore`, so I'm adding that one, and I'm adding one for a reduction, and might as well add firebase.

timneutkens:
🙌 :shipit:

rauchg:
Tests added in:
https://github.com/zeit/ncc/pull/24

index.js
nik-garmash: Maybe I'm wrong, but I thought the idea of this package was to mimic the behavior of Node, which supports only .js, .json and .node. After all, the fact that ncc uses webpack is an implementation detail, it's not exposed to the end user and they should not expect any webpack-specific features.

Btw, ".node" probably also needs some special treatment, even though it's used quite rarely.

timneutkens: Then I guess we have to force `.js` `.json` for webpack's resolve too 👍

timneutkens: I've updated the PR 👍

timneutkens:
Looks like firebase is solved by this PR 💯 

google-datastore still breaks because of grpc. I found this issue that was closed with nearly the same message: https://github.com/grpc/grpc/issues/11435

PR #26, opened 2018-11-20
Disable NODE_ENV replacement

timneutkens:
As per https://spectrum.chat/zeit/now/is-v2-doing-text-replacement-on-scripts~39d23a50-1650-4b6b-8698-da37888db457

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ module.exports = async (entry, { minify = true } = {}) => {
   const compiler = webpack({
     entry,
     optimization: {
+      nodeEnv: false,
       minimize: false
     },
     mode: "production",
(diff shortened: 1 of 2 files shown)


guybedford:
This seems like a bug in the NODE_ENV replacement itself to me, as the example uses it as an assignment target not a referenced identifier, which should be fixable in the replacement code itself.

TooTallNate:
Even still, we should have this feature disabled, since it would be including _build env vars_ into the resulting bundle, which are likely different than the _runtime env vars_.

timneutkens:
@guybedford yeah it seems like a bug in webpack's DefinePlugin.
And @TooTallNate is right, we don't want to do this in NCC.

SimenB:
Any chance of making this configurable? I wanna bundle a React app ([Ink](https://github.com/vadimdemedes/ink)), and it's _way_ faster in production mode

PR #27, opened 2018-11-20
Support fs static eval and encoding

guybedford:
Fixes https://github.com/zeit/ncc/issues/11.

Borrows the evaluation approach used by BrowserifyFS in https://www.npmjs.com/package/static-eval but doesn't use the other layers as we are working with ES modules instead of CommonJS so it is nice to keep the analysis at this level to continue to support these cases in ES modules and not just restrict these solutions to CommonJS files.


guybedford:
This has a regression on the integration test, just fixing that up.

guybedford:
CI failure seems unrelated to me (Yarn registry error?).

guybedford:
Will see if I can rework this for the webpack integration.

Issue #35, opened 2018-11-22
Go-style signature


rauchg:
The more I think about it, the more I'd like to support:

```
$ ncc run
```
and
```
$ ncc build
```

We would:
- [x] Add support for `pkg.main` / `index.js` if no argument is supplied
- [ ] `run` to build to a temporary location and clean it up on a best-effort basis
- [x] `build` to output to a default `dist` directory if `-o` is not supplied (in tandem with #30)

rauchg:
To elaborate on this:

>  clean it up on a best-effort basis

This would imply doing it on `beforeExit` or a similar process / signal hook.

rauchg:
Done

guybedford:
These are all completed.

PR #36, opened 2018-11-22
Directory-based builds / output 

guybedford:
This is an approach to #35 (although without the `run` command).

The API is changed to output `{ code, assets }`, where `assets` is an object of pathnames to buffers / strings (whatever webpack gives us from webpack asset emission).

The CLI is changed to output to `dist/index.js` by default including any assets.

I'm completely open to alternatives to any of this stuff... mainly just trying to get a base for fs relocations.


rauchg:
Excellent!

Issue #33, opened 2018-11-21
grpc bug


guybedford:
The issue with the package.json error there is the following line in node_modules/grpc/src/grpc_extension.js:

```js
var binding_path =
    binary.find(path.resolve(path.join(__dirname, '../package.json')));
```

where `__dirname` then has a different meaning in the built file than from the original.

the above `binding_path` is then passed directly to `require`. Ideally `binding_path` should be statically computed in its entitreity.

guybedford:
I've managed to get the package.json emitted as an asset and found... which then results in the following error:


    The gRPC binary module was not installed. This may be fixed by running "npm rebuild"
    Original error: Cannot find module 'C:\Users\Guy\Projects\ncc\test\tmp\assets\src\node\extension_binary\node-v64-win32-x64-unknown\grpc_node.node'

I think this means this issue is now aligned with binary support, given the pending updates to https://github.com/zeit/ncc/pull/39

Issue #17, opened 2018-11-16
FS inlining support


rauchg:
Lack of fs inlining breaks modules like:
- `saslprep`
- `pdfjs` (`fontkit` dep)

It yields a warning in `mongoose`, but that's due to `find-package-json`, which *I think* might not be statically analyzable[1]

[1] famous last words

timneutkens:
Sounds like this might work:
https://github.com/webpack-contrib/transform-loader
Combined with:
https://github.com/browserify/brfs

rauchg:
Let's make sure it works for this case:
https://github.com/zeit/ncc/issues/11

rauchg:
Fixed by #39

PR #39, opened 2018-11-22
Asset relocations

### LINKED ISSUE #2: `pdfkit` errors
ncc version: 0.22

Source:

```js
const fs = require("fs");
const PDFDocument = require("pdfkit");
const doc = new PDFDocument();
doc.fontSize(15).text("Hi there", 50, 50);
doc.pipe(fs.createWriteStream("out.pdf"));
doc.end();
```

Expected:

No output, working pdf:

```
▲  play-ncc/ node index.js
▲  play-ncc/
```

image

Received:

```
▲  play-ncc/ node out.js
internal/modules/cjs/loader.js:589
    throw err;
    ^

Error: Cannot find module 'iconv-lite'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:587:15)
    at Function.Module._load (internal/modules/cjs/loader.js:513:25)
    at Module.require (internal/modules/cjs/loader.js:643:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/Users/rauchg/Projects/play-ncc/out.js:1:270)
    at Module._compile (internal/modules/cjs/loader.js:707:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:718:10)
    at Module.load (internal/modules/cjs/loader.js:605:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:544:12)
    at Function.Module._load (internal/modules/cjs/loader.js:536:3)
```

Compilation output:

```
▲  play-ncc/ ncc index.js -o out.js
'iconv-lite' is imported by node_modules/fontkit/index.js, but could not be resolved – treating it as an external dependency
'iconv-lite' is imported by node_modules/restructure/src/DecodeStream.js, but could not be resolved – treating it as an external dependency
'iconv-lite' is imported by node_modules/restructure/src/EncodeStream.js, but could not be resolved – treating it as an external dependency
'iconv-lite' is imported by commonjs-external:iconv-lite, but could not be resolved – treating it as an external dependency
Circular dependency: node_modules/pdfkit/js/object.js -> node_modules/pdfkit/js/reference.js -> node_modules/pdfkit/js/object.js
Circular dependency: node_modules/pdfkit/js/object.js -> node_modules/pdfkit/js/reference.js -> commonjs-proxy:/Users/rauchg/Projects/play-ncc/node_modules/pdfkit/js/object.js -> node_modules/pdfkit/js/object.js
Circular dependency: node_modules/pdfkit/js/font.js -> node_modules/pdfkit/js/font/standard.js -> node_modules/pdfkit/js/font.js
Circular dependency: node_modules/pdfkit/js/font.js -> node_modules/pdfkit/js/font/standard.js -> commonjs-proxy:/Users/rauchg/Projects/play-ncc/node_modules/pdfkit/js/font.js -> node_modules/pdfkit/js/font.js
Circular dependency: node_modules/pdfkit/js/font.js -> node_modules/pdfkit/js/font/embedded.js -> node_modules/pdfkit/js/font.js
```

guybedford:
This implements https://github.com/zeit/ncc/issues/17 emitting assets as separate files.

It is based to the sourcemap branch as I was using that for debugging, but we can always rebase to master too.

It's a bit of a trick, which I'm not sure if anything else does, which is basically to detect references to `__dirname` or `__filename` then to statically analyze outward from those to see if it ends up forming a statically analyzable string literal to a known file path. In the process it will support path functions etc in this computation thanks to the browserify fs static-eval lib.

If it does find a file, it emits it as an asset of the build and relocates that static expression to reference the build output asset folder. So it's almost slightly lower level than the brfs approach in that we don't even need to analyze if there was an `fs.readFile` etc.

This implements #17, and fixes #2 as well as the saslprep bug.

The grpc bug in #33 is still failing unfortunately the issue is that we have:

```js
var binding_path =
    binary.find(path.resolve(path.join(__dirname, '../package.json')));
```

being successfully rewritten by this loader into:

```js
var binding_path =
    binary.find(__dirname + "/assets/package.json");
```

but the issue is that `binding_path` is then passed to `require(binding_path)` where Webpack will then rewrite that require to its own require, which throws during runtime.

One hack might even be to rewrite the runtime webpack require to fallback to using the local require instead of failing outright... will see if I can implement something along those lines tomorrow.


guybedford:
Note this is currently failing the self-build as webpack loaders aren't currently supported by ncc itself.

Currently looking into options.

guybedford:
The first flaw in this approach was that we don't want to emit "js" files as assets, which this trick would naturally do. The reason for this is that we might then be requireing an unbuilt file, with all its local requires etc... which wouldn't work.

I've just added a fix such that we only emit assets when the static file is not a ".js" or ".mjs" file, leaving those fully to webpack to handle.

guybedford:
I finally got the self-build running again here.

The final approach was to build the asset-relocator.js file separately in the build script, and then ensure that the __webpack_require__ within the build output will always fallback to the Node require in the build before throwing a not found error by hooking into the context module output.

PR #43, opened 2018-11-23
Fix top level return

rauchg:
It seems @sokra doesn't expose any clean hooks for extending the config, so for now we overload the method.

This makes the `path-platform` integration test pass.


timneutkens:
I think we need a test for this 🤔

TooTallNate:
@timneutkens The test is in https://github.com/zeit/ncc/commit/0a4c78989b548e524140f9f9f2eb4a11a243e0bc

vitalyster:
It is broken again since 0.8.0

vitalyster:
@guybedford [Here it is](https://github.com/vitalyster/ncc-gh43)

guybedford:
@vitalyster the problem there is that `return` is not supported in ES modules, and that won't work under `node --experimental-modules` either.

vitalyster:
But it works under Babel and works in ncc 0.7

guybedford:
That it ever works is the incorrect behaviour.

vitalyster:
Awesome! Work again in 0.14! Please don't break :)

PR #52, opened 2018-11-25
Add code coverage

guybedford:
Coverage is configured to run against the original sources, not the self-build to ensure the files are correctly filtered and applied.

I will see if I can increase the unit test coverage against this now for the cases in asset-inliner.


guybedford:
I've also updated the readme here - removing the inlining note and also adding a new caveats section covering what I think is the main part of the edge cases of the approaches.

Issue #49, opened 2018-11-25
Add subcommands to `ncc help`


rauchg:
Furthermore, get rid of `--help` and `--version` in favor of `ncc help` and `ncc version`.

PR #53, opened 2018-11-25
CLI usage updates

guybedford:
Adds `ncc help`, `ncc version`. Let me know if this captures everything for #49.

Also updates the README usage.


guybedford:
I've also added `ncc run` support here as well as ensuring package.json main support for the input (fixing #35)

PR #54, opened 2018-11-25
Simple binary approach

guybedford:
One of the major issues with binaries is that they are typically built for the target platform on install. This makes it very difficult to "statically" make binaries work for all platforms.

This PR brings back the google datastore test and gets it to pass by simply leaving the base-level binary package as an external of the build.

It may be possible to auto-detect binaries in this way (package.json with a `binary` field seems sufficient), and then inform users to keep them as dependencies so that this provides a fully automated binary workflow (combined with the corresponding node_modules location and build of the specific binary file on the target machine).

If the target architecture is known, then binary inlining could certainly be provided though. Most likely forking node-pre-gyp in such specific implementation work would be the best way to achieve this.

(note the reason this external approach works is because the previous asset relocation work overrides the webpack require to fall back to the build-time require).

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "prepublish": "npm test"
   },
   "devDependencies": {
+    "@google-cloud/datastore": "^2.0.0",
     "@sentry/node": "^4.3.0",
     "acorn": "^6.0.4",
     "apollo-server-express": "^2.2.2",
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -62,7 +62,7 @@ for (const integrationTest of fs.readdirSync(__dirname + "/integration")) {
   // ignore e.g.: `.json` files
   if (!integrationTest.endsWith(".js")) continue;
   it(`should evaluate ${integrationTest} without errors`, async () => {
-    const { code, map, assets } = await ncc(__dirname + "/integration/" + integrationTest, { sourceMap: true });
+    const { code, map, assets } = await ncc(__dirname + "/integration/" + integrationTest, { sourceMap: true, externals: ["grpc"] });
     module.exports = null;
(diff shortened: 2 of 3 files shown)


rauchg:
Very cool. I wonder if in this case we should also emit `node_modules`?

rauchg:
CCing @TooTallNate who might have thoughts about this

TooTallNate:
As far as `node-gyp` based dependencies goes, the presence of a `binding.gyp` file and a `build` dir in the dep signifies a native dependency.

I think when a native dep is encountered, we should `grep build/Release/*.node build/Release/*.so` and copy those into `<outdir>/build/`. Native deps that rely on `bindings` module would then _just work_, and for the cases that directly require the `.node` file we can rewrite the location to reference the location relative to the outdir.

This would handle _most_ cases I think. However I could imagine there being issues if there are conflicting binary artifacts in the outdir (for example, I know I name it `binding.node` in at least a few of my modules, as opposed to `<modulename>.node` which I think is more common).

TooTallNate:
Is this PR actually related to binary support?

guybedford:
It just gets a binary integration test to pass, but let's try and do better certainly. Happy to close this to tackle the more meaty approaches.

Zertz:
FWIW, while there might be a better or more robust solution, getting `"@google-cloud/datastore"` to work would be a win 😃

guybedford:
Note this is done with test options (setting grpc as an external) over being automatic or introducing code improvements. So it's really just adding a new unit test!

Zertz:
Is it possible to set externals within a deployment on Now or otherwise get grpc to work while a full fix is developed?

PR #59, opened 2018-11-26
Add CLI run example

guybedford:


rauchg:
I merged by I think "executable" might be confusing – perhaps "build output" ?

PR #61, opened 2018-11-26
Sourcemaps support

guybedford:
This adds a `--source-map` / `-s` flag for generating source maps through the CLI.

In addition, when used with `ncc run`, source maps support is automatically injected into Node with a build of source-map-support/register.

In the API I changed the input option from `sourceMap: true` to `sourcemap: true` as it seemed more consistent. Happy to change this back if anyone disagrees. Also added a `filename` option to the API as the source map needs a filename to reference to.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -43,7 +43,7 @@ require('@zeit/ncc')('/path/to/input', {
   minify: true, // default
   // externals to leave as requires of the build
   externals: ["externalpackage"],
-  sourcemap: true // default
+  sourceMap: true // default
 }).then(({ code, assets }) => {
   console.log(code);
   // assets is an object of asset file names to sources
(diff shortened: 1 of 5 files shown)


rauchg:
Question: how expensive are source maps? In other words, is there a case for not enabling them for the prod output?

readme.md
```diff
@@ -31,10 +31,10 @@ $ ncc build input.js -o dist
 Outputs the build of `input.js` into `dist/index.js`.
 
 ```bash
-$ ncc run input.js
+$ ncc run input.js --source-map
```
rauchg: This is a redundant example right? Since `run` will have it by default.

Per my comment below, I'm also considering whether `build` will have it as well, and we'll just go with `--no-source-map`.

guybedford:
Hmm, I just tested the average time it takes the tests to run with vs without sourcemaps and it was 67s with and 61s without.

I guess I'm just used to builds where source maps times dominate build times... but that certainly isn't the case here.

I've added a commit to set the default to use sourcemaps.

We could still stick with having them off for build by default, although that would need a separate `--source-map` flag.

Issue #47, opened 2018-11-25
Clean up README


rauchg:
- [x] Add badges
  - [x] CI
  - [x] Code coverage
- [x] Update usage info / API

guybedford:
Can this be closed?

PR #78, opened 2018-11-28
Add badges, update readme API options

guybedford:
This adds the CI and coverage badges.

I've added the coverage call to the CI, but the token and setup may need some checking here first.


guybedford:
Fixes #47.

PR #80, opened 2018-11-29
[Failing] Add isomorphic-unfetch test

joecohens:
@guybedford we had this report by a user which made me realize that `isomorphic-unfetch` was failing. I'm leaving you this pr with the failing test as an issue, please let me know if you need more information. 

screen shot 2018-11-29 at 1 25 02 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "got": "^9.3.2",
     "graphql": "^14.0.2",
     "ioredis": "^4.2.0",
+    "isomorphic-unfetch": "^3.0.0",
     "jest": "^23.6.0",
     "jimp": "^0.5.6",
     "koa": "^2.6.2",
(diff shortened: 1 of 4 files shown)


guybedford:
I've pushed a fix to this branch.

This was happening because:

* `node_fetch` has a "main": "lib/index" entry
* In `lib` there is both an `index.js` and an `index.mjs`
* We've disabled the "module" field, but we still have ".mjs" as an extension
* So Webpack was resolving "index.mjs", which exposed a `{ default: fetch }` API instead of `fetch` (ESM interop!)

Ideally we could just disable index.mjs in Webpack, but I wasn't sure how to do this.

So the alternative fix was just to have `.js` take precedence over `.mjs`.

rauchg:
Very nice fix @guybedford

joecohens:
Thanks @guybedford!

Issue #83, opened 2018-11-30
require.main.filename not shimmed...


username-II:
require.main.filename not shimmed...

timneutkens:
Could you provide more context?

timneutkens:
Ok so it looks like something Node.js provides by default:

https://nodejs.org/api/modules.html#modules_accessing_the_main_module

username-II:
Sorry, yes as shown in: https://nodejs.org/api/modules.html#modules_require_main
as described here: https://www.stefanjudis.com/today-i-learned/node-js-entry-script-with-process-mainmodule-or-require-main/
require.main.filename is handy for finding the main entry script filename, path and associated info.

rauchg:
It's very similar to #21

PR #84, opened 2018-11-30
Require main support

guybedford:
Supports treating "require.main" as a passthrough value so that it is the application-level `require.main` in the output.


guybedford:
Fixes #83.

Issue #65, opened 2018-11-26
Filenames of resources must be unique to avoid a crash


baboons:
Considering the examples for NCC with Apollo and an abstract structure with schemas as resources and sharing the same filenames. This will cause NCC to compile the first/last file only, and throwing an error. 

```
schemas\
      blogs\
           index.js
           schema.gql
      posts\
           index.js
           schema.gql
```

Output

```
   3kB  /schema.gql
6127kB  /index.js
```

At the moment this can only be resolved by renaming one of the schemas and make sure the file names are unique:

`const typeDefs = String(fs.readFileSync(__dirname + "/schema2.gql"));`

Output

```
   1kB  /schema2.gql
   3kB  /schema.gql
6127kB  /index.js
```

rauchg:
I'll close until we hear from @baboons whether this is happening on 0.3.0+

PR #85, opened 2018-11-30
Binary require support

guybedford:
This implements the simple case of (3) in https://github.com/zeit/ncc/issues/57.

Thanks @TooTallNate for the suggestions on this one - was exactly as you described and using a similar approach to node-native-loader.


guybedford:
This also fixes another bug where asset name deduping was not being shared between modules, which should resolve #65.

rauchg:
@guybedford has conflicts

guybedford:
Rebased.

PR #92, opened 2018-12-03
When.js failing test

guybedford:
When defines itself using the wrapper pattern:

```js
(function(define) { 'use strict';
define(function (require) {

	var timed = require('./lib/decorators/timed');
	var array = require('./lib/decorators/array');
	var flow = require('./lib/decorators/flow');
	var fold = require('./lib/decorators/fold');
	var inspect = require('./lib/decorators/inspect');
});
})(typeof define === 'function' && define.amd ? define : function (factory) { module.exports = factory(require); });
```

Similarly to https://github.com/zeit/ncc/pull/87 we have the static analysis problem of not knowing that the internal requires are external requires... (although in theory this static analysis is a trivial compiler problem :P).

This seems to be the bug behind "jugglingdb" builds as far as I can tell.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "twilio": "^3.23.2",
     "vue": "^2.5.17",
     "vue-server-renderer": "^2.5.17",
-    "webpack": "^4.26.0"
+    "webpack": "^4.26.0",
+    "when": "^3.7.8"
   }
 }
--- a/src/loaders/relocate-loader.js
+++ b/src/loaders/relocate-loader.js
@@ -50,7 +50,82 @@ function isExpressionReference(node, parent) {
 	return true;
 }
 
-const relocateRegEx = /_\_dirname|_\_filename|require\.main|node-pre-gyp|bindings/;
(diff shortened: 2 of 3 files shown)


guybedford:
One suggested fix for the above would be to "statically detect" these AMD-style wrappers, and "unwrap" them in a loader before passing back to Webpack for analysis.

rauchg:
I was just about to create a `jugglingdb` failing test, but it's this! :)

guybedford:
I've added the fix for this case, based on a very specific wrapper detection.

While the code is highly overly-specific, I think there will be ways to generalize the approach to better analysis methods as it expands, but i think as a starting point starting specific then moving to more general analysis makes sense, driven by the test cases.

rauchg:
@guybedford can you:
- fix the conflict
- write up an issue with the more general solution description
?

guybedford:
Posted https://github.com/zeit/ncc/issues/95 which would replace a lot of the logic with a partial evaluator.

PR #93, opened 2018-12-03
Binary relocations

guybedford:
Implements https://github.com/zeit/ncc/issues/57 in statically analyzing `require('bindings')` and `require('node-pre-gyp').find` loads and handling them as asset relocations.

Cases supported:

* `import binary from 'node-pre-gyp'; binary.find(__dirname + 'package.json')`  (plus `import *` variant)
* `require('bindings')('asdf')`
* `var bindings = require('bindings'); bindings('asdf')`

This should be merged with the other PRs, so will rebase depending on what order the merges happen.

Because of the issue with architecture matching, I originally wanted to enable this specifically behind a "binary" flag, but since the default case is then effectively for binaries just not to work, I've left this out for now.


guybedford:
Fixes #33.

PR #96, opened 2018-12-03
Fix source maps

rauchg:


src/sourcemap-register.js
```diff
@@ -0,0 +1,3 @@
+// note: this file is overriden by `scripts/build`
+// and substituted for the production release
+module.exports = require("source-map-support/register");
```
rauchg: This makes it work both inside `dist/` and in `src/`

guybedford: Ah, thanks for sorting out this use case too.

PR #98, opened 2018-12-03
Add segment test

guybedford:
This brings through the (passing) segment test form the more-tests branch.

With this merged, all the current tests on the more-tests branch will be in master.


guybedford:
Ahh, rebase fail - this is already in!

PR #100, opened 2018-12-04
Support Node shebang scripts

### LINKED ISSUE #69: File with shebang doesn't work
The first script I tried ncc on failed like this:

```
(node:71332) UnhandledPromiseRejectionWarning: Error: Hash: c93a3aedc91c7aab0770
Version: webpack 4.26.0
Time: 127ms
Built at: 2018-11-27 16:04:22
 Asset     Size  Chunks  Chunk Names
out.js  4.7 KiB       0  main
Entrypoint main = out.js
[0] ./2-stats2boundaries.js 1.11 KiB {0} [built] [failed] [1 error]

ERROR in ./2-stats2boundaries.js
Module build failed (from (webpack)/ncc/asset-relocator.js):
SyntaxError: Unexpected character '#' (1:0)
    at Parser.module.exports.pp$4.raise (/Users/stevebennett/npm/lib/node_modules/@zeit/ncc/dist/ncc/asset-relocator.js:16125:13)
...
```

The first line of that script is:

`#!/usr/bin/env node  --max-old-space-size=4096`

guybedford:
Fixes #69.

### DIFF
--- a/test/unit/shebang/input.js
+++ b/test/unit/shebang/input.js
@@ -1,2 +1,2 @@
-#!/bin/user/env
+#!/bin/user/env node
 module.exports = 'asdf';
--- a/test/unit/shebang/input.js
+++ b/test/unit/shebang/input.js
@@ -1,0 +1,2 @@
+#!/bin/user/env
+module.exports = 'asdf';
--- a/test/unit/shebang/output.js
+++ b/test/unit/shebang/output.js
@@ -1,0 +1,95 @@
+module.exports =
+/******/ (function(modules) { // webpackBootstrap
+/******/ 	// The module cache
+/******/ 	var installedModules = {};
+/******/
+/******/ 	// The require function
+/******/ 	function __webpack_require__(moduleId) {
+/******/
+/******/ 		// Check if module is in cache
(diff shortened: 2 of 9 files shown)


test/unit/shebang/input.js
```diff
@@ -0,0 +1,2 @@
+#!/bin/user/env
```
TooTallNate: `#!/bin/user/env` -> `#!/bin/user/env node` ?

test/unit/shebang/output.js
```diff
@@ -0,0 +1,95 @@
+module.exports =
```
TooTallNate: Should the shebang be copied to the output entrypoint file?

TooTallNate: And also possibly setting executable permissons (or copy the permissons from the entrypoint)?

guybedford:
I've added the features of #107 to this PR, without making the shebang explicit in the API and instead just passing it through in the code source as suggested.

rauchg:
@guybedford can you re-base please?

guybedford:
Rebased.

PR #101, opened 2018-12-04
Add MIT license

### LINKED ISSUE #97: Add License
It would be good to ensure this is included.

Any thoughts re which variant to use here?

guybedford:
Fixes #97.


refack:
GO Beavers!
image

PR #102, opened 2018-12-04
Add AMD unit test

### LINKED ISSUE #82: An amd test proposal
Let's add this snippet as a test for https://github.com/zeit/ncc/commit/f1eec46f7958564fe9f095894947e2aa09ed8057
```
console.log(
  ('function' === typeof define) && (define.amd != null)
);
```
Compiled javascript must return `false`. Without the commit, compiled snippet crashes.

guybedford:
This unit test ensures that AMD patterns are not detected, and AMD requires are not detected.

Resolves #82.


rauchg:
Absolute unit

PR #105, opened 2018-12-04
Minification support

### LINKED ISSUE #62: Minification
Currently, `minimize: true` being turned on is failing for an unknown reason with all integration tests.

We should be mindful here to only enable optimizations that are deemed safe for most Node.js programs and npm modules (including those that perform runtime introspection or reflection).

The pros / cons of this have to be carefully evaluated in contrast with what #51 provides.

guybedford:
Fixes #62.

Finally found the right combo of minification options to not get things to break here.

I would have used the Webpack integration of minification, but auth0.js has a *very* strange bug where terser just returns "undefined" for the source code, which needed a custom workaround (basically just to fallback to unminified when this happens).

This does slow down the tests quite a bit, but I think it is worth it to ensure minification is properly tested. I've left minification off by default though in builds for this reason.


rauchg:
@guybedford needs rebase

PR #106, opened 2018-12-04
Asset directory emission, test refactoring

### LINKED ISSUE #58: Asset wildcards and directory emission
Currently the static analysis of assets only supports emitting "files". Whenever there is a static expression that exactly corresponds to a file, that file is emitted.

Many dynamic loading cases use dynamic directory emission though:

```js
exports.getThing = function (name) {
  return fs.readFileSync(__dirname + '/things/' + name);
}
```

Now one thing we could do here is adjust the asset emission to support directory expressions as well.

This requires supporting wildcards in static analysis - when `name` is encountered, basically treating it as a wildcard string path and not stopping but continuing the static analysis. So in the above analysis when `/path/to/things/*` is formed as an expression, we see that it matches real files and then emit the folder with numeric deduping, along with all files matching the wildcard such that any loads within that folder will work out.

It gets a bit tricky, but it is very much possible to do sensibly.

This would fix super important real-world use cases like cowsay.

### LINKED ISSUE #68: Issues with `strong-globalize`
I'm using [Loopback](https://loopback.io/doc/en/lb3/) to build my project. Loopback/Stroploop internally uses **Globalize** and **cldr** for globalization in [strong-globalize](https://www.npmjs.com/package/strong-globalize). Looks like they use webpack while building.

And I'm getting the following error:
```
16753kB  /index.js                 [12444ms]
internal/modules/cjs/loader.js:589
    throw err;
    ^

Error: Cannot find module 'cldr'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:587:15)
    at Function.Module._load (internal/modules/cjs/loader.js:513:25)
    at Module.require (internal/modules/cjs/loader.js:643:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/tmp/86f4f7db49348/index.js:77424:18)
    at __webpack_require__ (/tmp/86f4f7db49348/index.js:21:30)
    at /tmp/86f4f7db49348/index.js:90591:4
    at Object.<anonymous> (/tmp/86f4f7db49348/index.js:90598:2)
    at __webpack_require__ (/tmp/86f4f7db49348/index.js:21:30)
    at Object.<anonymous> (/tmp/86f4f7db49348/index.js:306120:18)
```

guybedford:
Fixes #68.

This extends asset emission to direct emission of directories, fixing the strong-globalize test case.

I also went ahead and refactored the tests to use the "run" approach over an embedded eval. This simplifies the source maps handling but did require rewriting out the module.exports execution. Top-level promises work out here fine though since as a separate process, each test runs to completion.

This does not close #58 as it is effectively just the first step of that.


gijo-varghese:
@guybedford I'm getting the error with the latest version. I've updated in  https://github.com/zeit/ncc/issues/68

PR #112, opened 2018-12-05
v8cache experimental flag

### LINKED ISSUE #51: V8 Compile Cache
The idea is that we can emit not just `.js` files, but also the v8 code cache blobs.
And unlike `v8-compile-cache`, we could benefit from making all bootups fast, including the first one. Also unlike it, we can use the new `createCachedData` APIs which have some benefits.

- [ ] Make sure we disable the optimization gracefully when invoking in old Node versions
- [ ] Make sure to add a flag to disable it (`--no-v8-cache`)

More:
- https://github.com/nodejs/node/pull/20300
- https://github.com/nodejs/node/commit/4f67c6f667fa98e25ba311d3459fc8e270407250
- https://github.com/Microsoft/vscode/issues/62643
- https://nodejs.org/dist/latest-v10.x/docs/api/vm.html#vm_script_createcacheddata
- (legacy) https://nodejs.org/api/vm.html#vm_new_vm_script_code_options `produceCachedData`
- (legacy) https://www.npmjs.com/package/v8-compile-cache

guybedford:
Fixes #51.

Usage is `ncc build x.js --v8-cache` or the `v8cache: true` option for the API.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -10,10 +10,13 @@ async function main() {
   const { code: cli, assets: cliAssets } = await ncc(
     __dirname + "/../src/cli",
     {
+      filename: "cli.js",
       externals: ["./index.js"],
-      minify: true
+      minify: true,
+      v8cache: true
     }
   );
+
   const { code: index, assets: indexAssets } = await ncc(
     __dirname + "/../src/index",
     {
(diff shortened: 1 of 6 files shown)


guybedford:
Rebased.

While the build files look pretty ugly, this improves the self-build startup time by around 100ms on my machine.

guybedford:
The wrapper being used here was conflicting with Jest's Node hooks, so I've updated to a more robust wrapper approach.

PR #115, opened 2018-12-05
WIP: Per-source minification (DO NOT MERGE)

guybedford:
Here's a proof of concept towards caching for minification in https://github.com/zeit/ncc/issues/46.

Because minification is done per-source, the output currently looks like:

```js
module.exports =
/******/ (function(modules) { // webpackBootstrap
/******/ 	// The module cache
/******/ 	var installedModules = {};
/******/ ETC
/************************************************************************/
/******/ ([
/* 0 */
/***/ (function(module, exports, __webpack_require__) {
// minified line:
var bind=__webpack_require__(4),isBuffer=__webpack_require__(5),toString=Object.prototype.toString;
/***/ }),
/* 2 */
/***/ (function(module, exports, __webpack_require__) {

// another minified line:
var utils=__webpack_require__(0),normalizeHeaderName=__webpack_require__(26),DEFAULT_CONTENT_TYPE={"Content-Type":"application/x-www-form-urlencoded"}
// ETC
/***/ })
/******/ ]);
//# sourceMappingURL=index.js.map
```

It could be possible to hack the various webpack outputs to reduce the gaps here, but it definitely would be quite a hack.

The benefit at the end of the day would be per-source minification caching which would significantly speed up builds with minification.

Let me know if this seems like a worthwhile direction to continue along.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -137,13 +137,13 @@ module.exports = async (
             }
           }]
         },
-        {
+        ...minify ? [{
           test: /\.(js|mjs|tsx?)$/,
           use: [{
             loader: __dirname + "/loaders/terser-loader.js",
             options: { sourceMap }
           }]
-        }
+        }] : []
       ]
     },
     plugins: [


sokra:
This might be an interesting source if you looking for performance vs size tradeoffs too: https://github.com/terser-js/terser#terser-fast-minify-mode

guybedford:
@sokra good point, added in #161. Am I right in thinking Webpack already does the simple optimizations which would explain why we're seeing absolutely no difference?

PR #117, opened 2018-12-06
Add failing loopback test

guybedford:
With strong-globalize fixed, there is still a remaining bug in loopback reported by @gijo-varghese.

The test is failing with the following error:

```js
/private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T/39e4e63c55ba1/webpack:/node_modules/strong-globalize/lib/helper.js:80
    assert(validPath, '*** setRootDir: Intl dir not found under: ' + rootPath.toString());
^
AssertionError [ERR_ASSERTION]: *** setRootDir: Intl dir not found under: /private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T
    at Object.setRootDir (/private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T/39e4e63c55ba1/webpack:/node_modules/strong-globalize/lib/helper.js:80:1)
    at Function.SetRootDir (/private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T/39e4e63c55ba1/webpack:/node_modules/strong-globalize/lib/strong-globalize.js:62:1)
    at Object.<anonymous> (/private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T/39e4e63c55ba1/webpack:/node_modules/loopback/lib/globalize.js:10:1)
    at __webpack_require__ (/private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T/39e4e63c55ba1/webpack:/webpack/bootstrap:19:1)
    at Object.<anonymous> (/private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T/39e4e63c55ba1/webpack:/node_modules/loopback/lib/server-app.js:7:1)
    at __webpack_require__ (/private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T/39e4e63c55ba1/webpack:/webpack/bootstrap:19:1)
    at Object.<anonymous> (/private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T/39e4e63c55ba1/webpack:/node_modules/loopback/lib/loopback.js:12:1)
    at __webpack_require__ (/private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T/39e4e63c55ba1/webpack:/webpack/bootstrap:19:1)
    at Object.<anonymous> (/private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T/39e4e63c55ba1/webpack:/node_modules/loopback/index.js:11:1)
    at __webpack_require__ (/private/var/folders/v0/cjlw1c6d2_zfqp3v0pqshd3m0000gn/T/39e4e63c55ba1/webpack:/webpack/bootstrap:19:1)
```

This seems to still effectively be a bug in strong-globalize in that its dynamic algorithm is trying to load an internationalized folder we weren't able to detect.

This is a tricky one - will put some thought to it.


guybedford:
I've added two fixes here to get loopback to work:

1. There was a bug with directory asset emission stopping further assets from being emitted, which will be an important fix to get out.
2. I relaxed the restriction on directory asset emission to emit the nested directories and also to emit base-level package directories. The reason for this is that `loopback/intl` is expected to be available, but we only know from static analysis that it is accessing something in the `loopback` folder so we emit the whole thing.

(1) is uncontroversial.

(2) on the other hand will increase the output directory size quite a bit as it will potentially emit entire packages from node_modules. We could potentially look at ways to cut this back with various filtering heuristics (for example, hardcoding that any directory called "intl" should be emitted). But this really depends on if the added dist footprint is going to be a problem.

I'd suggest running tests for (2) to check how it affects output sizes and if that is going to be a problem or not.

If we want to wait on (2) as well, I can separate the fix from (1) in a different PR to get things moving with that in the mean time.

scripts/build.js
```diff
@@ -69,7 +69,7 @@ async function main() {
   // copy webpack buildin
   await copy(
     __dirname + "/../node_modules/webpack/buildin/*.js",
-    __dirname + "/../dist/buildin/"
+    __dirname + "/../dist/ncc/buildin/"
```
guybedford: In the process, the buildins directory got moved... I think this is because __dirname is being renormalized, but I didn't track this cause completely.

src/loaders/relocate-loader.js
```diff
@@ -558,28 +557,26 @@ module.exports = function (code) {
             transformed = true;
             magicString.overwrite(staticChildNode.start, staticChildNode.end, replacement);
           }
-          staticChildNode = staticChildValue = undefined;
         }
-        else if (stats && stats.isDirectory() &&
-            // dont emit __dirname or package base
-            staticChildValue !== path.dirname(id) &&
-            staticChildValue !== pkgBase) {
```
guybedford: This is the way in which directory emission is being widened - we were previously not emitting packages or __dirname references at all.

What this means is that code as simple as - `console.log(__dirname)` will cause the current directory to be emitted.

src/loaders/relocate-loader.js
```diff
@@ -558,28 +557,26 @@ module.exports = function (code) {
...
+        else if (stats && stats.isDirectory()) {
           let replacement = emitAssetDirectory(path.resolve(staticChildValue));
           if (replacement) {
             transformed = true;
             magicString.overwrite(staticChildNode.start, staticChildNode.end, replacement);
           }
-          staticChildNode = staticChildValue = undefined;
         }
+        staticChildNode = staticChildValue = undefined;
```
guybedford: This was the bug (1) in the directory emission that was stopping further directory emissions from working after passing the line above.

test/unit/asset-fs-inline-path-shadow/output.js
```diff
@@ -96,7 +96,7 @@ console.log(fs.readFileSync(__dirname + '/asset.txt', 'utf8'));
 
 (function () {
   var join = () => 'nope';
-  console.log(fs.readFileSync(join(__dirname, 'asset.txt'), 'utf8'));
+  console.log(fs.readFileSync(join(__dirname + '/asset-fs-inline-path-shadow', 'asset.txt'), 'utf8'));
```
guybedford: This is a consequence of directory emission widening - because there is a `__dirname` reference that can't be analyzed to anything more specific than that, the entire directory is emitted. Note how in theory it would enable the use case the unit test is eluding to.

guybedford:
I managed to put together some numbers for this from the integration tests:

* analytics-node - no assets, unaffected
* apollo - no assets, unaffected
* auth0 - no assets, unaffected
* aws-sdk - no assets, unaffected
* axios - no assets, unaffected
* azure-cosmos - no assets, unaffected
* azure-storage - no assets, unaffected
* bugsnag-js - no assets, unaffected
* firebase-admin - 10,049kB -> 10,155kB
* strong-globalize - 3,701kB -> 6,800kB
* pdfkit - 2573kB -> 2573kB
* jimp - 1068kB -> 1943kB

Looking at these numbers more closely it seems like we might actually be good to go on this one...!

rauchg:
This looks great. I think over-fitting is fine when FS stuff is involved, as to always prioritize that the programs work, first and foremost.

PR #122, opened 2018-12-06
Loopback PR redux

guybedford:
The previous PR that was merged was based against the bug fix it built on top of to pass... here's it back against master.


guybedford:
Note this is still needed to finish the merge from #117.

PR #138, opened 2018-12-09
Use bundled typescript

guybedford:
The issue fixed here is that the typescript loader loads typescript using a dynamic require in https://github.com/TypeStrong/ts-loader/blob/master/src/compilerSetup.ts#L15.

This is fixed using the previous dynamic require technique of allowing webpack to support a numeric `require` where `compiler` is simply the numeric `require.resolve('typescript')` replacement provided by Webpack.

We do this by patching the empty context module to attempt to load that numeric id when it can.

With this fixed, the remaining bug was that TypeScript loads its own types from the `lib` directory, but with logic that isn't statically analyzable by ncc, something like the following:

```js
processRootFile(ts.combinePaths(defaultLibraryPath, libFileName), /*isDefaultLib*/ true, /*ignoreNoDefaultLib*/ true);
```

where the asset emission would work if the above was `path.join(__dirname, 'lib', libFileName)` but doesn't work here because we can't statically analyze either `defaultLibraryPath`, `libFileName` or `ts.combinePaths`.

The work-around here was just like the webpack builtins - to manually relocate the `lib` folder into the ncc dist folder, which surprisingly then all seems to work out ok.

It would be worth testing it further though with various configurations to be sure...


package.json
```diff
@@ -79,14 +79,15 @@
     "rollup-pluginutils": "^2.3.3",
     "rxjs": "^6.3.3",
     "saslprep": "^1.0.2",
+    "season": "^6.0.2",
```
timneutkens: Is this a typescript dep?

guybedford: Ah, thanks that was left over from another branch - removed.

PR #151, opened 2018-12-13
Add support for tsconfig `paths`. (fixes #150)

### LINKED ISSUE #150: Support for tsconfig `paths`
In my `tsconfig.json` I use the `paths` option:

```json
"compilerOptions": {
  [...]

  "paths": {
    "@*": ["./src/*"]
  },
}
```

Which allows me to import files as such:

```ts
import validateToken from "@auth/jwt"; // => <root>/src/auth/jwt.ts
```

However ncc's module resolution is incompatible with this feature, giving the following error:

```bash
ncc: Module directory "<dir>" attempted to require "@auth/jwt" but could not be resolved, assuming external.
``` 

Something like `tsconfig-paths-webpack-plugin` could be used to support this configuration.

*I could submit a PR for this if there are no objections.*

odensc:
This PR adds support for the `paths` option in tsconfig, using `tsconfig-paths-webpack-plugin`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ dist/**/*.js
 !test
 !test/integration
 !test/integration/*.js
+!test/integration/*.ts
+!test/unit
+!test/unit/**
 !dist/
 !dist/ncc/
 !dist/buildin/
--- a/package.json
+++ b/package.json
@@ -85,6 +85,8 @@
     "terser": "^3.11.0",
     "the-answer": "^1.0.0",
     "ts-loader": "^5.3.1",
+    "tsconfig-paths": "^3.7.0",
+    "tsconfig-paths-webpack-plugin": "^3.2.0",
     "twilio": "^3.23.2",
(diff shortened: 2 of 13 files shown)


rauchg:
@odensc I really cannot accept this without tests. Any chance you can add a few? Thank you!

rauchg:
I believe we already have a typescript test in there! `request-ts.ts`, which tests bundling `request`

odensc:
@rauchg I couldn't actually find that file in the repo - until 804ee91e1e00a1ba8615bbc6bb33b4e0c6caf79c there were no `.ts` files at all. I've added a couple tests though, let me know how that looks. :)

rauchg:
My bad @odensc – I'd totally forgotten to push my `.ts` integration test:
https://github.com/zeit/ncc/commit/d4d0eadfa1298a661e3447110e2f48f64d3af91d

Would you mind adding another one for paths?

Issue #50, opened 2018-11-25
Add examples folder


rauchg:
- [ ] Hello world
- [ ] Multiple bundles (e.g.: `cli.js` -> `dist/cli/` and `main.js` -> `dist/main`)

macklinu:
Would also be good to add examples for programmatic use and TypeScript support 😎 

Can get started on a PR for these

maxbeatty:
What's the optimal way to do multiple bundles? Programmatic?

Would there ever be a day where `ncc` takes a glob to build? (e.g. `ncc build *.js` or `ncc build {cli,main}.js`)

macklinu:
@maxbeatty looks like there is a little bit of talk about supporting multiple entry points in #113. As of now, I believe the only way to support multiple bundles right now would either be:

```sh
ncc build a.js -o dist
ncc build b.js -o dist
```

or with a custom build script using the ncc Node API:

```js
const path = require('path')
const ncc = require('@zeit/ncc')

// Whatever options you want. This is an example.
const options = {
  minify: true,
  sourceMap: true,
  watch: false,
}

Promise.all([
  ncc(path.resolve('./a.js'), options),
  ncc(path.resolve('./b.js'), options),
])
  .then(([a, b]) => {
    // Builds succeeded!
    // You can write the code and sourcemaps to disk or whatever you'd like to do here.
    console.log(a.code)
    console.log(a.assets)
    console.log(a.map)
  })
  .catch((err) => {
    // There was an error in one of the builds.
    console.error(err)
  })
```

(where `a.js` and `b.js` live in the folder you are running the script from. Those sorts of file paths would need to be updated to suit your needs.)

I think examples for programmatic use (both for one entry point and multiple entry points) would be welcome additions to this repo to help guide people in understanding what options are available / best practices are around multiple bundles + programmatic use. 👍

guybedford:
Examples were added in https://github.com/zeit/ncc/commit/c813b07096f2d3e89dae9a3d0917413a2c19ba24.

We can reopen if we need to add any new examples.

PR #153, opened 2018-12-14
Create hello-world example

macklinu:
This is a start at #50. The main goal of this PR is to figure out the primary structure for an example directory (README.md structure, how to install `@zeit/ncc`, how to communicate which commands to run, etc.) so that it becomes easier to add other examples moving forward. Feedback welcome (inline comments incoming)!


examples/hello-world/index.js
```diff
@@ -0,0 +1 @@
+console.log("Hello World!");
```
macklinu: This is the simplest Hello World program I could think of, but if there's something else we'd like to express with the Hello World program, I'm happy to change this.

examples/hello-world/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "devDependencies": {
+    "@zeit/ncc": "file:../../"
```
macklinu: I'm installing ncc from the root of the repo - thoughts?

examples/hello-world/README.md
```diff
@@ -0,0 +1,17 @@
+# Hello World
+
+## How to use
+
+Within this directory, run `yarn` to install dependencies.
+
+To execute the Node.js program, run:
+
+```sh
+npx ncc run index.js
```
macklinu: We could also put these commands in `package.json#scripts`:

### `package.json`

```json
{
  "scripts": {
    "execute": "ncc run index.js",
    "build": "ncc build index.js -o dist"
  }
}
```

### `README.md`

```md
To execute the Node.js program, run `yarn execute`.
To build the Node.js program, run `yarn build`.
```

I chose [npx](https://www.npmjs.com/package/npx) as a way to execute `ncc` from within the `examples/hello-world` directory, but open to feedback on what would be helpful/simple.

examples/hello-world/package.json
rauchg: Sounds good!

examples/hello-world/index.js
rauchg: Embedding a dependency would be nice. For example, `chalk`

examples/hello-world/README.md
rauchg: I like the idea of `scripts` better. `build` and `dev` IMO.

macklinu:
@rauchg addressed your comments. Here are examples of running the npm scripts from the `examples/hello-world` directory:

image

image

examples/hello-world/package.json
macklinu: One issue I'm noticing here is that if I run `yarn install` in the `examples/hello-world` directory, it can take ~2 minutes to install `@zeit/ncc` from two directories above.

If I change the version from `file:../../` to `latest` and install from npm, it downloads much faster.

[I notice Next.js' examples use `latest`](https://github.com/zeit/next.js/blob/cd1d3640a9abfa3df3cadf712fe095aa5326a089/examples/using-preact/package.json#L11), and perhaps that's a pattern we could adopt here. Or maybe there's a way to improve the speed of installing a file dependency with yarn using [a command flag](https://yarnpkg.com/lang/en/docs/cli/install/). If you have suggestions about next steps with this, let me know. 😄

rauchg:
Thanks @macklinu

examples/hello-world/package.json
rauchg: Let's do `latest` @macklinu. If we install all the dev deps from the directory above, it'll take a very long time. We have all those deps for integration testing purposes only

rauchg: @macklinu ping

macklinu: I addressed this in de48c58, but I don't think GitHub automatically resolved this thread. 😅

macklinu:
@rauchg this should be a good for another review - apologies for not sending a comment after my latest commit to communicate that more clearly. Let me know if I can help out without anything else on this PR. Will keep contributing example PRs if we like the start of this format. 👍 

(This PR is also good to be squashed and merged unless you'd like me to squash/rebase it manually.)

rauchg:
Awesome @macklinu ! Next step would be to show off an example of using `arg` and a CLI with subcommands, where each subcommand is a different bundle

PR #158, opened 2018-12-15
feat: auto-install dependencies

### LINKED ISSUE #56: Dependency installation
When I use `go` and `go.mod`, installing deps is a side-effect of execution or building. 

```
▲  golang-example/ (master) cd hello/
▲  hello/ (master) ls
.               ..              go.mod          hello.go
▲  hello/ (master) go run hello.go
go: finding rsc.io/quote v1.5.2
go: finding rsc.io/sampler v1.3.0
go: finding golang.org/x/text v0.0.0-20170915032832-14c0d48ead0c
go: downloading rsc.io/quote v1.5.2
go: downloading rsc.io/sampler v1.3.0
go: downloading golang.org/x/text v0.0.0-20170915032832-14c0d48ead0c
Hello, world.
```

I think `ncc` should do the same, with the ability to opt-out.

fmiras:
Closes #56


src/cli.js
```diff
@@ -11,12 +14,13 @@ Commands:
...
-  -S, --no-source-map   Skip source map output
-  -e, --external [mod]  Skip bundling 'mod'. Can be used many times
-  -q, --quiet           Disable build summaries / non-error outputs
-`;
+  -o, --out [file]        Output directory for build (defaults to dist)
+  -M, --no-minify         Skip output minification
+  -S, --no-source-map     Skip source map output
+  -e, --external [mod]    Skip bundling 'mod'. Can be used many times
+  -q, --quiet             Disable build summaries / non-error outputs
+  -D, --no-dependencies   Disable autoinstall dependencies feature
```
rauchg: `--no-dep-install` perhaps?

rauchg:
It would be cool to:
- Run `yarn` if `yarn.lock` is present
- Do it as a webpack plugin so that we can auto-install on a per-`import` basis

In other words, if you `import` a file from some other folder that has a `package.json`, might be worth to run `yarn` there as well.

fmiras:
Added `yarn.lock` support, pending port to webpack plugin

rauchg:
Actually @fmiras I realized we shouldn't do that, because it breaks the principles of doing what "node does". In other words, if we want to run `yarn` "in other directories", we should just use `dependencies` to define a module with a `file` prefix.

So, I think your original patch is great, if we support both `yarn` and `npm`

fmiras:
Sorry @rauchg but I didn't get it. What is the thing that we shouldn't do? The port it as webpack plugin or running `yarn` as default if `yarn.lock` present? In this case we already support both `yarn` and `npm`.

rauchg:
Closing due to lack of activity for now

PR #161, opened 2018-12-16
Separate minify compress option

guybedford:
This adds a separate `compress` / `--compress` CLI and API option, that need to be explicitly opted into for terser compression.

The size of the ncc build itself is exactly the same with and without Terser compression, yet takes _27s_ with compression, vs _25.7_ without compression.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -61,6 +61,8 @@ file is necessary. Most likely you want to indicate `es2015` support:
 ```js
 require('@zeit/ncc')('/path/to/input', {
   minify: true, // default
+  // use minification compression
+  compress: false,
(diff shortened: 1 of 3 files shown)


guybedford:
Given the lack of any effect of terser compression, it could even be worth disabling it completely and not even having a flag here.

I've added another commit that does that, which can be reverted if this is really needed, but I don't see a good reason to include it personally since it seems Webpack is already doing the most basic optimizations for us anyway.

guybedford:
(When I mean that the build size is exactly the same, I really do mean it is identically _5,150,515_ bytes between the two)

Issue #46, opened 2018-11-25
Cache


rauchg:
We need a simple yet effective strategy to make several invocations of `ncc build` or `ncc run` in a row re-use as much work as possible.

- [ ] FS cache for the CLI
  - Uses the most appropriate location for each OS (like yarn)
  - Flag to ignore it for both `run` and `build` (`--no-cache`)
  - Subcommands to find it and clean it (`ncc cache dir` and `ncc cache clean`)
- [ ] In-memory interface for the programmatic API (`ncc()`)

guybedford:
Of the time spent in the build, over 50% is now spent on minification, with only a very small fraction of time spent in the relocation loader.

Unfortunately Webpack internals don't seem like they offer much in the way of persistent caching right now (correct me if I'm wrong, but got this from - https://github.com/webpack/webpack/issues/6527).

So I think the initial focus needs to be on minification (and TypeScript support when merged).

So the question then is how to properly cache minification in a way that can be shared between builds.

One option here could be running minification as a loader, having it apply per-input-source before Webpack does its tracing and concatentation. This would rely on caching all the other loaders and their emissions as well (including the relocation loader) and then having the previously generated source used when the input matches.

I'm not sure that per-source minification is a proven approach (please let me know if I'm wrong), so this would be yet to be seen what wins can be made here.

guybedford:
This can be considered completed given that we are now running the Webpack v5 cache with these options enabled.

PR #162, opened 2018-12-16
Webpack next caching

guybedford:
Implements the "next" webpack branch caching for #46.

Here are the stats for the self build:
* Without cache: 31s
* With cache: First build 32s, Subsequent builds 15s

And in fact the remaining time there is actually mostly minification, so it should be possible to get this down to of the order of seconds pretty easily if we go with the approach in https://github.com/zeit/ncc/pull/115.

The Webpack upgrade was mostly pretty smooth. There was an issue with the parser injection for the top-level return handling, which was tricky to fix as there seemed to be a different instance of the parser in use in Webpack to the one we could hook into.

Instead I have disabled that test for now, and also posted a Webpack bug and PR for the root cause at https://github.com/webpack/webpack/issues/8509 and https://github.com/webpack/webpack/pull/8510.

#### Cache Handling

I couldn't work out how to manage the cache with a custom persistence layer here, so Webpack is entirely owning the file system side of things here.

There is currently no cache eviction, this seems like it will be an area of development in Webpack 5 to handle evicting old items as the cache reaches a certain size based on usage, otherwise it it at risk of eventually "taking over" the file system entirely :P

The other thing that was mentioned in the PR thread was handling "relocations" of the cache. Eg if I have app1/node_modules/dep1 and another app has the same app2/node_modules/dep1 there is a lot of duplication going on. Again, this is something that would likely need to happen on the Webpack core side to better handle these types of redundancies. Perhaps this relocation is more of a problem for ncc maintaining its own cache, as webpack actually defaults to `node_modules/.cache/webpack` for a per-project cache, while the ncc cache is across-projects.

The final thing to watch out for will of course be invalidation issues, which will likely be the most important thing to test for releasing this.

#### API

The API gets a `cache` option which is the string directory to use for the cache defaulting to `os.tmpdir/ncc-cache`. The cache can be disabled with `cache: false` (it is on by default).

#### CLI

The CLI commands are `ncc cache clean`, `ncc cache dir` and `ncc cache size`.

It also comes with a new `--no-cache` flag.

### DIFF
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -8,7 +8,7 @@ for (const unitTest of fs.readdirSync(`${__dirname}/unit`)) {
   it(`should generate correct output for ${unitTest}`, async () => {
     const testDir = `${__dirname}/unit/${unitTest}`;
     const expected = fs
-      .readFileSync(`${testDir}/output.js`)
+      .readFileSync(`${testDir}/output${global.coverage ? '-coverage' : ''}.js`)
       .toString()
       .trim()
       // Windows support
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -60,6 +60,9 @@ for (const integrationTest of fs.readdirSync(__dirname + "/integration")) {
   // disabled pending https://github.com/zeit/ncc/issues/141
   if (integrationTest.endsWith('loopback.js')) continue;
 
+  // TODO: Reenable this test once https://github.com/webpack/webpack/issues/8509 is fixed
+  if (integrationTest.endsWith('path-platform.js')) continue;
+
   it(`should evaluate ${integrationTest} without errors`, async () => {
     const { code, map, assets } = await ncc(
       __dirname + "/integration/" + integrationTest
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -60,9 +60,6 @@ for (const integrationTest of fs.readdirSync(__dirname + "/integration")) {
   // disabled pending https://github.com/zeit/ncc/issues/141
   if (integrationTest.endsWith('loopback.js')) continue;
 
-  // TODO: Reenable this test once https://github.com/webpack/webpack/issues/8509 is fixed
-  if (integrationTest.endsWith('path-platform.js')) continue;
-
   it(`should evaluate ${integrationTest} without errors`, async () => {
     if (global.gc) {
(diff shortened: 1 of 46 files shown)


guybedford:
CI failure here is due to Javascript heap out of memory :(

I've tried bumping the heap size.

rauchg:
Seems like it could be related to a memory leak in Webpack 5. I would refrain from bumping it further, as we shouldn't see such increased memory use relative to the webpack 4 branch

guybedford:
So it looks like this is actually only happening when running the tests with coverage (as is done on CI).

Seems like it might be https://github.com/facebook/jest/issues/5837 then.

guybedford:
Ok, finally got this fixed with a manual GC run. Turns out the bump before would have fixed it to, it was just done on the wrong test.

For the coverage - because we don't have coverage on the `cli.js` file, adding anything to this code decreases the overall percentage considerably. I can try and expose an API for the cli.js internals so we can call that programatically and get this back up. Created https://github.com/zeit/ncc/issues/163.

test/index.test.js
```diff
@@ -56,10 +56,25 @@ function clearDir (dir) {
...
+      console.log(`GC Completed, Heap Size: ${process.memoryUsage().heapUsed / 1024 ** 2} MB`);
+    }
+
     const { code, map, assets } = await ncc(
       __dirname + "/integration/" + integrationTest,
-      { minify: true, sourceMap: true }
+      {
+        // enabling cache stops saslprep.js test from working
+        // which seems like a Webpack 5 cache bug...
+        cache: false,
```
guybedford: The `saslprep.js` test was failing until I disabled the cache explicitly. The failure was that the asset wasn't being emitted.

When the saslprep.js test is run on its own in isolation it passes fine, regardless of cache state, so it seems like some kind of interaction between webpack cache and previous build memoization.

Because it works in isolation fine, and this is the only test that fails, I don't think this is as bad as it sounds, but it's worth bearing in mind.

guybedford:
The remaining error here is that switching from the "src" to the "dist" ncc build results in Webpack assigning different ids to the modules so the unit test output tests fail!

For now, I've just commited the two versions as I have no idea how to trace down why there is a difference between the two in Webpack 5 only...

PR #169, opened 2018-12-17
Webpack next per-source minification

guybedford:
This implements #115 on top of #162, handling minification per-source, and therefore also caching minification per source.

The previous results it turns out were actually without minification due to #168. Here are the numbers with minification:

The build time summary of the self build here is:
* With cache, with minification before this PR: First build 64s, Subsequent builds 47s
* With cache, with minification after this PR: First build 64s, Subsequent builds 36s

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -36,6 +36,10 @@ async function main() {
     __dirname + "/../src/loaders/shebang-loader"
   );
 
+  const { code: terserLoader, assets: terserLoaderAssets } = await ncc(
+    __dirname + "/../src/loaders/terser-loader"
+  );
+
   const { code: tsLoader, assets: tsLoaderAssets } = await ncc(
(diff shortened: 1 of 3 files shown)


guybedford:
Here's the difference in footprint on the self-build:

* 2,691,197 bytes with per-source minification
* 2,456,293 bytes with whole-file minification

the difference there is basically the wrappers not being minified - `function (module, exports, __webpack_require__)` sort of thing instead of `function(a,b,c)`. There's no easy safe way to hack in the minification of the wrappers though unfortunately.

So the question then is the extra 10s on the self build (plus inability to parallelize) worth 100KB? I'd argue it might not be...

rauchg:
@guybedford I agree it's not. Keep in mind that once we investigate #51 I'm not even sure *minification is worth it to begin with*

Issue #114, opened 2018-12-05
Loader parallelization for performance


guybedford:
The standard way to improve performance would be to run loaders on separate threads.

There should be various ways to do this while retaining the synchronization of the `assets` shared object through a custom channel.

Issue #41, opened 2018-11-23
Supporting require.resolve dynamic passing


guybedford:
Within the use cases around custom loaders (think Babel plugins, webpack loaders), there are a number of edge cases of dynamic require that come up. While Webpack can get quite far in computing dynamic requires like `require(require.resolve('./asdf.js'))`, there is a tricky case where there is a separation between the resolution and the require:

```js
// example use case something like "plugin: require.resolve('./asdf')" being passed as an argument
const req = require.resolve('./asdf.js');
require(eval('"' + req + '"')); // or any other untracable logic before passing to require
```

In this case, what Webpack does is replace the require.resolve part with the ID of the module in the webpack bundle, so we get something like:

```js
const req = 234;
__webpack_require__(123)(eval('"' + req + '"'))()
```

Where **123** is effectively an inlined "throwing" module which will give the error "Module 234 not found".

The above then fails as a module not found, all of the time.

The naive fix I was thinking to implement would to instrument the "throwing module" (123 in the example) to first check the __webpack_require__ cache for the numeric ID.

TooTallNate:
This is a webpack-specific hack that works https://github.com/TooTallNate/node-bindings/commit/40dcc1ffad085e22cf4b730a002ceeed77297ac5

guybedford:
Nice, that seems a sensible approach to getting bindings working. Although it relies on the node_modules location right?

TooTallNate:
It's just a way to "break out" from the webpack "fails as a module not found, all the time" scenario and get access to the native `require`. Indeed this is a runtime thing though so maybe it's not applicable here.

guybedford:
@TooTallNate another option for breaking out of webpack without per-package hacks might be to reinstrument https://github.com/webpack/webpack/blob/master/lib/ContextModule.js#L610 through the plugin system (I'm sure there's a way!) to check the global require on not found first.

guybedford:
I tried hacking the above approach to make webpack loaders work, but unfortunately passing string ids through the resolver was causing breaks in the internal loader logic (eg `loader resolve -> "1026"` kind of thing).

That was the primary goal of this issue, although the approach may still work for Babel plugins.

Leaving this one for now.

guybedford:
Reopening this issue, as coming up with a solution to these dynamic cases will solve #141, #148.

guybedford:
[...]

PR #174, opened 2018-12-17
WIP - Per-source threaded minification [Do not merge]

guybedford:
This is some experimental work on parallelizing minification.

It doesn't currently support the self-build due to the good old dynamic require issue (#41), so its pending a workaround for that.

So I'm running this with - `node ./src/cli.js build twilio.js -o twilio`.

When building twilio I get the following numbers:
* 37s minified on master
* 27s per-source minification, parallel as in this PR
* 21s unminified

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "stripe": "^6.15.0",
     "terser": "^3.11.0",
     "the-answer": "^1.0.0",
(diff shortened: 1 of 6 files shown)


guybedford:
Fixes #114.

rauchg:
Not huge on minifying for the time being, so let's leave this for later

PR #175, opened 2018-12-18
Disable minification

guybedford:
This disables minification by default, significantly improving build times.

Instead the `--minify` and `--sourceMap` flags are now opt-in.

The self-build still retains minification.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -66,14 +66,14 @@ file is necessary. Most likely you want to indicate `es2015` support:
 
 ```js
 require('@zeit/ncc')('/path/to/input', {
-  // provide a custom cache path or disable caching
-  cache: "./custom/cache/path" | false,
   // externals to leave as requires of the build
   externals: ["externalpackage"],
   minify: false, // default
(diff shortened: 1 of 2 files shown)


macklinu:
As a user of build/bundling tools, I think this change makes sense. I usually expect to opt into a production mode or minification mode, so if this also helps improve build times by default, then that's great too. 🙌

rauchg:
Why source maps too?

guybedford:
@rauchg source maps seems important when minified, but note that Node.js doesn't actually natively support source maps (without the source-map-support third-party injection) so that generating them by default seems unnecessary. It only cuts off about a second of build time on a 28 second large build, but that seems worth it to me.

TooTallNate:
Do you think it would make sense to include and setup `source-map-support` in the output bundle implicitly when `--sourceMap` is specified.

guybedford:
@TooTallNate that's a good suggestion, I've created https://github.com/zeit/ncc/issues/180.

PR #179, opened 2018-12-20
Undocument watcher

guybedford:
If we're still not sure about including the watcher I'd suggest just undocumenting it in the mean time.

It's useful to maintain the implementation as it provides a great benchmark for the caching work.


sokra:
I might make sense to add a note that you need to `yarn add chokidar` to use the watcher, otherwise the dependency is missing.

guybedford:
Do we still want to do this?

rauchg:
@guybedford Let's leave it documented for now, and let's aim to have watch-like perf for re-runs :)

PR #182, opened 2018-12-20
Write assets with exec permissions

### LINKED ISSUE #172: Permission denied on vendor binary
Running `ncc run input.js` returns `.../vendor/primitive-darwin-x64: Permission denied`.

The binaries in `node_modules/sqip/vendor/...` are executable, but the binaries copied to the ncc output seem to lose the executable flag.

### input.js
```javascript
const sqip = require("sqip");
const result = sqip({ filename: "./input.jpg" });
```

Is it possible to keep the executable flag in the ncc output?

### LINKED ISSUE #198: Access rights don't seem to be preserved when copying files
it seems like there’s a bug in the file copy logic, it doesn’t preserve access rights:

That’s the original file:

```
-rwxr-xr-x  1 timneutkens  staff  103908468 Dec 27 15:32 thefile
```

Copied into the dist folder:

```
-rw-r--r--  1 timneutkens  staff  103908468 Dec 27 15:45 thefile
```

@guybedford I can provide a reproduction if needed, if you need one send me a DM on slack 👍

guybedford:
Fixes #172.
Fixes #198

This writes all assets with exec permissions, if we had a list of extensions to handle here we could do that too, but I wasn't sure what an exhaustive list would like like.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -200,7 +200,7 @@ switch (args._[0]) {
       for (const asset of Object.keys(assets)) {
         const assetPath = outDir + "/" + asset;
         mkdirp.sync(dirname(assetPath));
-        fs.writeFileSync(assetPath, assets[asset]);
+        fs.writeFileSync(assetPath, assets[asset], { mode: 0o777 });
       }
 
       if (!args["--quiet"]) {


rauchg:
I think for security reasons we should just *maintain* the exec permissions if they exist, right?

guybedford:
Good point... @sokra is there anyway to associate permissions with emitted files? Otherwise we can maintain a side table for this.

TooTallNate:
Ya the "stats" of the emitted files should be matched as closely as possibly I think.

src/cli.js
```diff
@@ -200,7 +200,7 @@ switch (args._[0]) {
       for (const asset of Object.keys(assets)) {
         const assetPath = outDir + "/" + asset;
         mkdirp.sync(dirname(assetPath));
-        fs.writeFileSync(assetPath, assets[asset]);
+        fs.writeFileSync(assetPath, assets[asset], { mode: 0o777 });
```
TooTallNate: Let's make sure to look up the existing permissions on the file, instead of blanket 777 permissions.

guybedford:
There's another issue here with caching actually and that is that webpack handles the asset emission on a cached build, but we have no idea where the asset came from in a cached build to apply the correct permissions as we don't even have the out-of-band permissions dictionary populated if we used such an approach.

This then means either disabling the cache unless there is a way for webpack to allow us to attach metadata to emitted assets, either indicating the original path or the permissions to use so that this information can be maintained.

@sokra would value your thoughts on this one.

sokra:
You can use the cache API to cache your metadata.

guybedford:
@sokra which cache API is this? I can't find any code examples or documentation or mentions in the v5 release notes.

guybedford:
The closest I can find is the mention to the `Cache` plugin in the release notes. Does this mean I should require this plugin directly and extend it or something? Or access its lifecycle through a hook?

TooTallNate:
Did this land? Or what's the status? I'm experiencing this issue with `ncc` v0.15.0.

guybedford:
This did land, with detailed handling to ensure the input permissions match the output permissions. Can you share a replication?

TooTallNate:
```
$ ls -l
total 8
-rwxr-xr-x 1 nrajlich staff  23 Feb 12 11:53 bootstrap.js*
-rw-r--r-- 1 nrajlich staff 118 Feb 12 12:34 t.js

$ ncc build t.js
0kB  dist/bootstrap.js
2kB  dist/index.js
2kB  [295ms]

$ ls -l dist/
total 8
-rw-r--r-- 1 nrajlich staff   23 Feb 12 12:38 bootstrap.js
-rw-r--r-- 1 nrajlich staff 1631 Feb 12 12:38 index.js

$ ncc version
0.15.0
```

guybedford:
@TooTallNate it turns out this was actually a cache issue, where the permissions cache needs to persist between builds. I did write the code for this, but didn't enable this as I hadn't yet seen it trigger. Enabled in https://github.com/zeit/webpack-asset-relocator-loader/commit/85bfce894d8c67124a5813fbbe0b27dd346fddf4.

PR #187, opened 2018-12-20
remove overwrite of inputFileSystem and use the default (cached) filesystem

sokra:
It's 20% faster for a cold build for me


guybedford:
Does the default implementation then provide the graceful-fs-like features for high levels of file access?

timneutkens:
@guybedford the default filesystem uses graceful-fs

timneutkens:
Specifically the creation of the input/output filesystem is found here: https://github.com/webpack/webpack/blob/master/lib/node/NodeEnvironmentPlugin.js#L14-L17

PR #189, opened 2018-12-21
Update webpack + pack mode

sokra:
* update webpack (for performance fixes)
* close the compiler even in case of errors
* switch to `pack` store mode (better performance)
* externals signature changed in v5

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,6 @@
 const resolve = require("resolve");
 const fs = require("graceful-fs");
+const crypto = require("crypto");
 const { sep } = require("path");
 const webpack = require("webpack");
 const MemoryFS = require("memory-fs");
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -54,8 +54,7 @@ module.exports = (
       type: "filesystem",
       cacheDirectory: typeof cache === 'string' ? cache : nccCacheDir,
       name: "ncc",
-      version: require('../package.json').version,
-      store: "instant"
```
guybedford: If switching to the `pack` mode, should we also now ensure that the cache is a per-project cache in the node_modules folder instead of the single global cache?

sokra: That would make sense.

Another option is to keep the cache location, but use a different `cache.name` for each directory.

Like this: `name: hashOf(directory)`

guybedford: The risk with the global cache with this new storage seems to be disk usage growth right?

I guess this can be mitigated with a simple eviction strategy in due course. But that was another argument for per-project caching.

The argument for global caching is that ideally it would be great if the global cache could be smart enough to know the hashes of the dependencies (yarn.lock again!), and thus share compilations.

Not sure which of the above you want to shoot for here, but will trust your judgement!

src/index.js
```diff
@@ -53,9 +62,8 @@ module.exports = (
     cache: cache === false ? undefined : {
       type: "filesystem",
       cacheDirectory: typeof cache === 'string' ? cache : nccCacheDir,
-      name: "ncc",
-      version: require('../package.json').version,
-      store: "instant"
+      name: `ncc_${hashOf(entry)}`,
```
guybedford: Should this not be a hash of the project directory, so that if the input changes the cache can still be used?

PR #194, opened 2018-12-26
add TypeScript example

maxbeatty:
add a simple TypeScript example ref #50


rauchg:
Thanks a lot @maxbeatty !!

Issue #192, opened 2018-12-24
ncc fails to build chromeless because mqtt SyntaxError: Unexpected character '#'


styfle:
Create a file called `index.js` which works fine when running `node index.js`.

```js
const { Chromeless } = require('chromeless');

async function getScreenshot(url) {
    const chromeless = new Chromeless();
    const filePath = await chromeless.goto(url).screenshot();
    console.log(filePath);
    await chromeless.end();
    return filePath;
}

getScreenshot('https://zeit.co').then(() => console.log('success')).catch(() => console.error('failed'));
```

Then try `ncc build index.js -o dist` and get the following error:

<details>
<summary>Click to show output</summary>
```
ncc: Module directory "/Users/styfle/Desktop/ncc-broken/node_modules/got" attempted to require "electron" but could not be resolved, assuming external.
ncc: Module directory "/Users/styfle/Desktop/ncc-broken/node_modules/ws/lib" attempted to require "utf-8-validate" but could not be resolved, assuming external.
ncc: Module directory "/Users/styfle/Desktop/ncc-broken/node_modules/ws/lib" attempted to require "bufferutil" but could not be resolved, assuming external.
Error: Hash: e6543104c891a74ef2bc
Version: webpack 4.26.0
Time: 3353ms
Built at: 12/23/2018 11:42:17 PM
       Asset      Size  Chunks  Chunk Names
    index.js  4.68 MiB       0  main
index.js.map  1.66 MiB       0  main
Entrypoint main = index.js index.js.map
  [3] external "fs" 42 bytes {0} [built]
  [7] external "util" 42 bytes {0} [built]
  [9] external "path" 42 bytes {0} [built]
 [18] external "os" 42 bytes {0} [built]
 [23] ./node_modules/chromeless/dist/src/util.js 36.1 KiB {0} [built]
 [52] ./node_modules/chromeless/dist/src/chrome/local.js 12 KiB {0} [built]
 [53] ./node_modules/chrome-remote-interface/index.js 1.18 KiB {0} [built]
[111] ./node_modules/chromeless/dist/src/queue.js 5.42 KiB {0} [built]
[112] ./index.js 371 bytes {0} [built]
[113] ./node_modules/chromeless/dist/src/index.js 422 bytes {0} [built]
[114] ./node_modules/chromeless/dist/src/api.js 11.9 KiB {0} [built]
[127] ./node_modules/chrome-launcher/dist/index.js 424 bytes {0} [built]
[...]

PR #196, opened 2018-12-27
Fixup shebang loader order

guybedford:
This ensures that analysis can still be correctly applied alongside the shebang loader, fixing the error in #192.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -113,12 +113,6 @@ module.exports = (
     module: {
       rules: [
         {
-          parser: { amd: false },
-          use: [{
-            loader: __dirname + "/loaders/shebang-loader.js"
-          }]
-        },
-        {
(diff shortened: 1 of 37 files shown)


guybedford:
I've also updated the Webpack fixtures here, as it seems the output module numbering has been changed by new work on the v5 branch.

guybedford:
This is quite odd, the unit tests were passing with the updates in the commit there, but now seem to have stopped working again. I'm not sure why the webpack output keeps changing like this perhaps @sokra has an idea? The tests are passing fine for me locally with this yarn.lock.

guybedford:
Ahh, the issue here was that there are two sets of unit tests - one with and without coverage. Completely forgot, despite being responsible for this :P

PR #201, opened 2018-12-31
add example with multiple bundles

maxbeatty:
adds example with multiple bundles (`main` and `cli`) using separate calls to `ncc build`. module is admittedly simple to keep focus on build process in `package.json`

ref #50


rauchg:
Epic! Thanks so much @maxbeatty

PR #207, opened 2018-12-31
add example with programmatic build

maxbeatty:
Adds an example programmatic build with multiple entry points. I tried to structure `scripts/build.js` so it could easily be repurposed for a single entry point.

ref #50


rauchg:
Amazing. Thanks again @maxbeatty

PR #210, opened 2019-01-09
Disable ts-loader typechecking for non-build files (resolves #208)

### LINKED ISSUE #208: Monorepo problem - Can't find `package`
Greetings!
I have searched for issues related to the one me and @mattapperson are facing, but I have failed to find it. Please advise if this is a known issue.

## Summary

Monorepo available [here](https://github.com/mattapperson/now-issue-example). We can successfully run it locally, but building with `cd app1 && ncc build index.ts` fails when executed for the first time. However, executing it again succeeds!

## Step By Step
```
git clone https://github.com/mattapperson/now-issue-example
cd now-issue-example
cd app1 && npm i
ncc build index.ts
ncc build index.ts
```

Any clues?

guybedford:
Up until now the ts loader has been running in the mode where it runs all typechecking (like what you get in the VScode console for all the types in the project). Because this feature doesn't naturally support webpack caching, disabling it is the easiest fix for #208.

I tend to think a more permissive build makes sense as well and I believe is in line with what Deno offers too.

The main concern with an alternative fix like @sokra describes is that it would mean always loading the ts loader upfront (therefore slowing down startup) unless we can get a hook into the cache usage to know if a ".ts" file is being loaded or not in order to conditionally do this (perhaps Tobias will know if this is possible).

But it seems easier and a better experience I think to just go for the one-liner :)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -131,6 +131,7 @@ module.exports = (
           use: [{
             loader: __dirname + "/loaders/ts-loader.js",
             options: {
+              transpileOnly: true,
               compilerOptions: {
                 outDir: '//'
               }


guybedford:
@sokra if we wanted to know if there is a ".ts" file loaded in the build to know to load the ts loader afterCompile hook for type checking (without incurring that cost for every load), is there a reliable hook that could get this information in time?

rauchg:
Favored #213 over this one

PR #211, opened 2019-01-09
Output asset permissions

guybedford:
This stores the permissions of assets in a side table and ensures the corresponding output asset permissions.

In theory this is a major change as the assets API is changed to be an object of `{ [assetName]: { source, permissions } }` instead of `{ [assetName]: source }` now.

@sokra an interesting thing here was it turned out unnecessary to maintain the cache as for some reason it seems like Webpack isn't actually caching the sources with emitted assets. Ideally these should be cached though I think?

I've left the cache persistence of asset permissions commented out for now given the above.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -23,8 +23,8 @@ const hashOf = name => {
 		.slice(0, 10);
 }
 
-const nodeLoader = require(__dirname + "/loaders/node-loader.js");
-const relocateLoader = require(__dirname + "/loaders/relocate-loader.js");
+const nodeLoader = eval('require(__dirname + "/loaders/node-loader.js")');
+const relocateLoader = eval('require(__dirname + "/loaders/relocate-loader.js")');
 
 module.exports = (
   entry,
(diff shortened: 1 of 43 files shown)


src/index.js
```diff
@@ -23,8 +23,8 @@ const hashOf = name => {
 		.slice(0, 10);
 }
 
-const nodeLoader = require(__dirname + "/loaders/node-loader.js");
-const relocateLoader = require(__dirname + "/loaders/relocate-loader.js");
+const nodeLoader = eval('require(__dirname + "/loaders/node-loader.js")');
+const relocateLoader = eval('require(__dirname + "/loaders/relocate-loader.js")');
```
sokra: Why not `require("./loaders/node-loader.js")` and define it as `external`

guybedford: I just tried this and it doesn't work (get `Cannot read "assets" of undefined` indicating there may be separate instances in play, or something else).

PR #213, opened 2019-01-09
Disable ts caching for typechecking

guybedford:
This is the alternative fix to #210 that retains the typing errors being emitted.

The approach here is to disable caching for ts compilations entirely to ensure that the types of the local files are checked everytime.

To utilize the webpack 5 cache for this work would likely require the ts-loader project to serialize its internal memoizations using the webpack cache api while itself becoming uncacheable. //cc @johnnyreilly

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1,5 +1,5 @@
 const ncc = require("../src/index.js");
-const { statSync, writeFileSync } = require("fs");
+const { statSync, writeFileSync, readFileSync } = require("fs");
 const { promisify } = require("util");
 const { relative } = require("path");
 const copy = promisify(require("copy"));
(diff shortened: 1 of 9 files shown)


guybedford:
(Note caching still applies fine to everything in node_modules of course)

johnnyreilly:
Hey @guybedford!

> To utilize the webpack 5 cache for this work would likely require the ts-loader project to serialize its internal memoizations using the webpack cache api while itself becoming uncacheable. //cc @johnnyreilly

I'm afraid I'm not too close to this issue; got a lot going on right now.

From your comment am I right in understanding that you think that making the changes suggested on #208 seem unwise?

https://github.com/TypeStrong/ts-loader/issues/894

Also, are you aware of https://github.com/Realytics/fork-ts-checker-webpack-plugin

This plugin works nicely in concert with ts-loader; allowing you to run ts-loader in `transpileOnly` mode whilst the plugin handles the typechecking. For performance reasons my default is always to use the `fork-ts-checker-webpack-plugin`

guybedford:
Thanks for the update @johnnyreilly. Our use case for the webpack cache is having the TS internal cache persisted between runs, which I don't think either project would offer us currently? Work towards that is what I was referring to in my comment. Running the separate projects with threading definitely seems like it would be a worthwhile perf improvement though only for the cold build times (and watcher I suppose).

johnnyreilly:
> Our use case for the webpack cache is having the TS internal cache persisted between runs, which I don't think either project would offer us currently? 

Totally correct yes

> Running the separate projects with threading definitely seems like it would be a worthwhile perf improvement though only for the cold build times (and watcher I suppose).

Yes you should definitely see a gain in those scenarios.

Thanks for giving me a heads up BTW; had no idea ts-loader was being used in NCC 😁

PR #218, opened 2019-01-15
Support -e flag in ncc run

guybedford:
This works by creating a symlink to the local node_modules folder in the temporary build location.


guybedford:
Fixes https://github.com/zeit/ncc/issues/154.

PR #232, opened 2019-01-17
[Do not merge] Revert "handle assets loaded with relative paths (#217)"

guybedford:
This reverts commit b5c9d1e1fa03c49b48bb34fed09d86522f9159a7.

It turned out while the tests passed, this broke the self build and caused issues with the TypeScript build too.

The result is that this approach simply isn't suitable for relative assets (a bottom up asset trigger), so we need to rather approach this asset case as top down (whitelist functions on assets like `fs.readFile` as opposed to detecting the asset expression itself, like we can do safely for __dirname etc).

This took me way to long to track down in bisect due to some weird cache and test interactions that were easily missed.


test/index.test.js
```diff
@@ -1,5 +1,5 @@
 const fs = require("fs");
-const ncc = global.coverage ? require("../src/index") : require("../src");
+const ncc = global.coverage ? require("../src/index") : require("../");
```
guybedford: Note this was the slip that hid the bug on master :)

guybedford:
Marked as do not merge, to rather merge this via https://github.com/zeit/ncc/pull/233.

rauchg:
Closed per #233

PR #233, opened 2019-01-17
Use the local TypeScript version when applicable

guybedford:
This also refactors the way that dynamic requires are handled by ensuring that `require(nonAnalyzableExpression)` will always defer to the `__non_webpack_require__`. This also allows us to remove the previous hack of the "Not found" context in Webpack doing a fallback to the environment require.

The internal loader loads are then altered to use a non-analyzable require to load dynamically in production. Note that this is the sort of feature code splitting would allow us to avoid as it would work naturally with the types of optimizations involved. So that it looks like a hack is simply because we aren't taking advantage of code splitting which would allow it not to be a hack (as in the whole ncc project would be one build as opposed to many then).

This is based to https://github.com/zeit/ncc/pull/232 which turned out to be quite a rabbit hole. This can be merged and that one closed as a result.


mikaelbr:
Think this might have introduced an issue (related #237) which also breaks typescript support for Now.sh with `@now/node@canary`. Sorry I haven't dug deep enough to find what is causing it.

mikaelbr:
Hypothesized a fix here after a limited analysis, but not on my pc to do proper debugging and verification: https://github.com/zeit/ncc/issues/237#issuecomment-458076300

PR #234, opened 2019-01-17
Asset relocator refactoring

guybedford:
This implements the refactoring described in https://github.com/zeit/ncc/issues/223.

Binding analysis is now fully general with no specific import / require / binding cases.

It's a pretty major refactoring, but seems to have worked out well in the end luckily!


guybedford:
(this is based to #232 so should be merged after #233).

PR #241, opened 2019-01-28
Use createCachedData correctly

sokra:
It need to be called after the script really ran.
v8 tracks type info while interpreting the script,
from these info compiled code is created.
Not executing functions doesn't track these info
and compiled code can't be created.

This change stores the new cached data everytime
the code runs, just before the process exit.
At time time most info is available.

This gives a 20%-30% performance boost for me

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -307,9 +307,11 @@ module.exports = (
       assets[filename + '.cache.js'] = code;
       if (map)
         assets[filename + '.map'] = map;
-      code = `const { readFileSync } = require('fs'), { Script } = require('vm'), { wrap } = require('module');\n` +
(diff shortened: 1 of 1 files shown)


guybedford:
Very cool, you've just invented the self-optimizing package :)

guybedford:
@sokra Im still reeling from this, this really is such a unique CS concept! A truly self-optimizing program. Profile-guided optimizing compilers be damned!

guybedford:
Also, in theory, since the cached optimization data is maintained transitively through every execution, there is also a convergence property in play here of optimizing the cached data over all executions thereby ending up with the best cache analysis, although that is depending on how v8 handles unused optimization data between runs.

PR #258, opened 2019-01-31
nodemon support

guybedford:
This adds nodemon support for the SIGUSR2 handler, as described in https://github.com/zeit/ncc/issues/253#issue-404488287.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -283,10 +283,14 @@ async function runCmd (argv, stdout, stderr) {
                 reject({ silent: true, exitCode: code });
               process.off("SIGTERM", exit);
               process.off("SIGINT", exit);
+              process.off("SIGUSR1", exit);
+              process.off("SIGUSR2", exit);
             }
             ps.on("exit", exit);
             process.on("SIGTERM", exit);
             process.on("SIGINT", exit);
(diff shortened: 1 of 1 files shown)


rauchg:
I'm ok with supporting this if we document in `README`? Unless it's a standard that I don't know about. cc @adamelliotfields

guybedford:
So as far as I'm aware Node.js has no "generic" signal handling - all signals need to be handled individually by name, which does seem quite restrictive unfortunately.

In theory, we could support all signals as exit indicators (http://man7.org/linux/man-pages/man7/signal.7.html and I seriously considered this), but a good approach I think can be to expand this list based on user scenarios like this one. Unless anyone has other suggestions for how best to handle these signal responses generically.

guybedford:
Another altenative here might be just to remove this error case scenario entirely and always just act as if `ncc run -f` were the default.

rauchg:
I hear you. I'm still really on the fence about watch mode. I'm not convinced `ncc` is as fast as it can be yet when re-running it "cold".

guybedford:
I've posted https://github.com/zeit/ncc/pull/265 which can supersede this.

PR #259, opened 2019-01-31
Disable run --watch

guybedford:
This adds a proper error message on the `ncc run --watch` case, which otherwise has some edge cases around signal handling that need to be worked out to fully support this. Some discussion on this is in https://github.com/zeit/ncc/issues/253, and the feature could certainly be developed further if desired.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -173,6 +173,9 @@ async function runCmd (argv, stdout, stderr) {
       if (args["--out"])
         errFlagNotCompatible("--out", "run");
 
+      if (args["--watch"])
+        errFlagNotCompatible("--watch", "run");
+
       outDir = resolve(
         require("os").tmpdir(),
         crypto.createHash('md5').digest(resolve(args._[1] || ".")).toString('hex')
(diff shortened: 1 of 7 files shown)


src/cli.js
```diff
@@ -311,19 +314,19 @@ async function runCmd (argv, stdout, stderr) {
...
   }
 
   function errTooManyArguments (cmd) {
-    nccError(`Error: Too many ${cmd} arguments provided\n${usage}`);
+    nccError(`Error: Too many ${cmd} arguments provided\n${usage}`, 2);
   }
 
   function errFlagNotCompatible (flag, cmd) {
-    nccError(`Error: ${flag} flag is not compatible with ncc ${cmd}\n${usage}`);
+    nccError(`Error: ${flag} flag is not compatible with ncc ${cmd}\n${usage}`, 2);
```
guybedford: I've extended the use of the `2` exit code here for CLI validation errors, as was provided for the help output already.

I'm not sure who implemented this code in the first place, but it seems a useful separation and helps to work with the code-based tests, but I can certainly remove it too and move this to an integration test rather as well.

PR #265, opened 2019-02-01
run -f as the default

guybedford:
This avoids the annoying message about left-over temporary build files on previous `ncc run` calls, and just always attempts to remove these files on `ncc run`.

This also avoids the need for the PR https://github.com/zeit/ncc/pull/258.


adamelliotfields:
Just pulled it down - working great with Nodemon so far.

Thanks @guybedford and @rauchg for the quick patch. I hope to contribute more in the future!

PR #294, opened 2019-02-25
Support fallback requires

guybedford:
This includes an upgrade to the asset relocator loader to fix https://github.com/zeit/ncc/issues/284.

The unit test case is included in the commit there - https://github.com/zeit/webpack-asset-relocator-loader/commit/e082f9b09574631cf64bf6de006b7e42d57b07e0.

A full integration test is still pending on this one. @Siyfion perhaps you can share a full example repo or test this out in your replication there?

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "@google-cloud/bigquery": "^2.0.1",
     "@google-cloud/firestore": "^0.19.0",
     "@sentry/node": "^4.3.0",
-    "@zeit/webpack-asset-relocator-loader": "0.1.10",
+    "@zeit/webpack-asset-relocator-loader": "0.1.12",
     "analytics-node": "^3.3.0",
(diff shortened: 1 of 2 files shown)


Siyfion:
@guybedford As I mentioned in the linked issue, an example repo that has the original issue is this:
https://github.com/prisma/graphql-yoga/tree/master/examples/prisma-js

And I can confirm that this _does_ fix it.

c316:
I have the same problem, how can I test this fix with my repo and this pull request? Is there a way to install the now cli from this pull request?

c316:
@guybedford any update here? I attempted to build now with this version of NCC, but I couldn't get it going. I'd be happy to help with a full reproduction if you have any pointers on getting this version of ncc into now-cli.

guybedford:
The update here is now effectively merged given the loader is at the latest. So this should come downstream in due course.

PR #301, opened 2019-03-06
Update jugglingdb to 2.0.1

styfle:
This fixes a potential security vulnerability.

Related to https://github.com/1602/jugglingdb/pull/460

Fixes: #303

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "isomorphic-unfetch": "^3.0.0",
     "jest": "^23.6.0",
     "jimp": "^0.5.6",
-    "jugglingdb": "^2.0.0-rc8",
+    "jugglingdb": "2.0.1",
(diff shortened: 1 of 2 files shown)


styfle:
> ✕ should execute "ncc run jugglingdb.js" (818ms)

@guybedford I broke it 😞 

I'm not sure what the original use case was here.

Should we continue testing the old version and add a new test for the new version? 🤔

rauchg:
Tests are failing

guybedford:
I've reported on this case here - https://github.com/zeit/ncc/issues/303.

guybedford:
I've just pushed an update to the asset relocator loader here, which goes with the approach (1) described in #303. This does have some minor perf cost (roughly 5%), but allows this case to be built and handled properly, ensuring we only build one copy of jugglingdb.

PR #307, opened 2019-03-14
Treat uglify-js as empty by default

guybedford:
This fixes the pug build, by always ignoring uglify-js builds. I know it doesn't sound ideal, but that really is the best fix we can come up with here given there isn't an analysis solution with the current technology.

When this happens the following warning is output:

```
ncc: Ignoring build of uglify-js, as it is not statically analyzable. Build with "--external uglify-js" if this package is needed.
```


rauchg:
I think this is a good solution for now, considering the popularity of the module

PR #311, opened 2019-03-14
Update browserify wrapper handling

### LINKED ISSUE #299: Error: Cannot find module 'http'
~https://github-code-review-notifier-1ev76qgm5.now.sh/_logs~

guybedford:
This includes the update to the asset relocator loader to improve the Browserify wrapper handling (https://github.com/zeit/webpack-asset-relocator-loader/commit/2d0e7c7e2cc4a2df2baf8803867f186cb4e5dce1).

Specifically we extend the browserify wrapper detection to be able to detect the uglification of the wrapper, as well as ensuring that the require detection is converted to apply to `__non_webpack_require__`.

Fixes #299.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "stripe": "^6.15.0",
     "terser": "^3.11.0",
     "the-answer": "^1.0.0",
+    "tiny-json-http": "^7.0.2",
     "ts-loader": "^5.3.1",
     "tsconfig-paths": "^3.7.0",
     "tsconfig-paths-webpack-plugin": "^3.2.0",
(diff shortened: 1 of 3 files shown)


styfle:
@guybedford Thanks! Can you add an integration test for `tiny-json-http` to prove that this fixes the bug?

guybedford:
The test is a unit test in the asset relocator loader. My hesitation with
adding full integration tests for everything is we end up with a 10 minute
test script, but I get the arguments for that too.

rauchg:
+1 on integration test

guybedford:
Sure, I've added an integration test here.

PR #314, opened 2019-03-17
Fixup tests on mac including asset symlinks support

guybedford:
This updates the webpack-asset-relocator-loader to support the tfjs-node build that was failing on MacOS.

The failure was due to shared libraries with .so extensions not begin emitted on mac, and also due to not having support for symlinked assets, as the pathing of the shared library relies on symlinks. Both of these are fixed in the latest webpack-asset-relocator-loader release at https://github.com/zeit/webpack-asset-relocator-loader/commit/23b0dbffce22fbe19f51c292a73f40ba82ab8386.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -259,6 +259,7 @@ module.exports = (
 
   function finalizeHandler () {
     const assets = Object.create(null);
+    const symlinks = Object.assign(Object.create(null), relocateLoader.getSymlinks());
     getFlatFiles(mfs.data, assets, relocateLoader.getAssetPermissions);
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -259,6 +259,7 @@ module.exports = (
 
   function finalizeHandler () {
     const assets = Object.create(null);
```
styfle: I'm curious, why use an object instead of Map?

guybedford: I guess I'm just used to exposing API output as JSON-like data?

PR #318, opened 2019-03-20
Only emit known symlinks

guybedford:
The new symlink handling that was added has to integrate with the cache, which means that there is a risk of previously cached symlinks being emitted when they are no longer needed.

To avoid this case, symlinks are filtered to only include symlinks to known assets.

~~I need to add a test for this still, but wanted to check this approach against CI first.~~ Coverage provided by Tensorflow and Sharp tests.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,7 @@
 const resolve = require("resolve");
 const fs = require("graceful-fs");
 const crypto = require("crypto");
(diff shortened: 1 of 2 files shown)


guybedford:
We've actually got coverage on this through the Tensorflow and Sharp tests, so this is good to go.

PR #325, opened 2019-03-26
Print version from programatic API

styfle:
This will print the ncc version and the file being compiled when using the programmatic API `const ncc = require('ncc')`. Previously, we were only printing the version when the CLI was used.

I also added `quiet` so that the programmatic API can enable/disable quiet mode.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -45,6 +45,7 @@ module.exports = (
 ) => {
   if (!quiet) {
     console.log('ncc: Version ' + nccVersion);
+    console.log('ncc: Compiling file ' + filename);
   }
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -38,9 +39,14 @@ module.exports = (
     sourceMap = false,
     sourceMapRegister = true,
     watch = false,
-    v8cache = false
+    v8cache = false,
+    quiet = false
```
guybedford: We should probably document this in the README.

styfle: Thanks, done! ✅

PR #327, opened 2019-03-28
Relative sourcemap base

### LINKED ISSUE #319: Support source-relative paths in source maps for debugger support
When running `ncc build --source-map --no-source-map-register` we output source maps that convert the `webpack:///file.js` paths into `/file.js`, but these still won't work with the Node inspector (even though sourcesContent is provided).

If we want to support these debugger workflows we could carefully ensure that the paths match up to the original source files relative to the dist folder. It will just be about working through the base pathing consistencies here.

guybedford:
This fixes #319, by ensuring that the source maps paths match up to the input folder, using a custom `sourceMapsBasePrefix` option.

Ideally the CLI should know to configure this option based on the output directory location relative pathing, but by default it will likely usually be `../`.

I compared the source maps here to the ones generated by TypeScript and the output file is pretty much identical now, but when I tried testing with `node --inspect-brk` in the Chrome dev tools, I still couldn't get the source map to apply.

I ran the source maps in the Webpack visualizer at https://sokra.github.io/source-map-visualization and can confirm they are all working 100%.

The only thing I can think here is either if my local setup isn't running file system source maps for some reason (so @styfle perhaps you can verify your side), or if the Webpack boostrap file is somehow making things invalid as it is the first source (`"/webpack/bootstrap"`). I tried changing this name to different forms but that didn't seem to make a difference.

Not sure what else to try, but source maps debugging is always like this the first time :)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -68,6 +68,7 @@ require('@zeit/ncc')('/path/to/input', {
   externals: ["externalpackage"],
   minify: false, // default
   sourceMap: false, // default
+  sourceMapBasePrefix: '../' // default treats sources as output-relative
   // when outputting a sourcemap, automatically include
   // source-map-support in the output file (increases output by 32kB).
   sourceMapRegister: true, // default
--- a/src/index.js
+++ b/src/index.js
@@ -38,7 +38,8 @@ module.exports = (
     sourceMap = false,
     sourceMapRegister = true,
     watch = false,
-    v8cache = false
+    v8cache = false,
+    sourceMapBasePrefix = '../',
   } = {}
 ) => {
   const resolvedEntry = resolve.sync(entry);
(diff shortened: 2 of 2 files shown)


guybedford:
There's a really odd CI failure on this one...

guybedford:
Ok I just tested in a TypeScript project and I'm also seeing the source maps not applying there so this may actually be something to do with my setup.

styfle:
I had the same problem with the inspector and Chrome in the past couple weeks. Maybe it’s a mismatch between v8 in Node vs Chrome?

Try VS Code and see if that works because the tsc source maps work for me with Code. 

I’ll restart CI.

styfle:
@guybedford This is really close!

I found that the source map is observed by VS Code, however the line numbers look wrong.


## Steps to reproduce

```sh
git clone https://github.com/styfle/ncc-bug-sourcemap
cd ncc-bug-sourcemap
git checkout pr327
```

### Working with tsc

```sh
yarn build
code . # open with VS Code
node --inspect-brk dist
# Visit the Debugger and select "Attach"
```

### Not working with ncc

```sh
# below is my copy of ncc built from your branch
/path/to/ncc/dist/ncc/cli.js build index.ts --source-map --no-source-map-register
node dist
code . # open with VS Code
node --inspect-brk dist
# Visit the Debugger and select "Attach"
```

animated gif

guybedford:
I just tested this out and I can replicate the source maps issue. But in my own test, the source maps worked fine.

I ran the source maps through the checker visualization at https://sokra.github.io/source-map-visualization/#custom and get them lining up correctly, see below.

image

So the above actually seems to be a VSCode / debugger source maps issue. This might be due to Node.js issues with source maps, which are being fixed thanks to the new compileFunction approach which has been worked on by bcoe.

PR #341, opened 2019-04-01
Temporary fix for pkg misbehaving when v8 cache is used

igorklopov:
Currently when v8 cache is used everything works well except `pkg` prints a message `Pkg: MAGIC_NUMBER_MISMATCH`. This is a temporary fix until i fix that message.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -323,11 +323,16 @@ module.exports = (
         assets[filename + '.map'] = { source: JSON.stringify(map), permissions: defaultPermissions };
         map = undefined;
       }
-      code = `const { readFileSync, writeFileSync } = require('fs'), { Script } = require('vm'), { wrap } = require('module');\n` +
+      code =
+        `if (process.pkg) {\n` +
+          `module.exports=require('./${filename}.cache.js');\n` +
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -323,11 +323,16 @@ module.exports = (
         assets[filename + '.map'] = { source: JSON.stringify(map), permissions: defaultPermissions };
         map = undefined;
       }
-      code = `const { readFileSync, writeFileSync } = require('fs'), { Script } = require('vm'), { wrap } = require('module');\n` +
+      code =
+        `if (process.pkg) {\n` +
```
guybedford: How are you setting `process.pkg`?

igorklopov: `pkg` sets this before running any user code. So one can check the mode

guybedford:
Also does this still happen with https://github.com/zeit/ncc/pull/241 reverted.

src/index.js
guybedford: Ahh, I see :)

styfle: 😄  https://github.com/zeit/pkg

Issue #342, opened 2019-04-01
--watch not working?


dpeek:
The watch flag doesn't seem to work for me.

To reproduce in empty dir:

```
yarn init -y
yarn add @zeit/ncc
echo "console.log('test');" >> index.js
ncc build index.js # works
ncc build index.js --watch # fails without error
```

Environment:

```
$ node --version
v11.13.0
$ ncc version
0.17.2
```

dpeek:
Works for me, thanks for the quick fix!

Side note: is there a way to get yarn link working with binaries (like ncc cli)? Would like to use this fix before you cut a new release.

guybedford:
@dpeek you're welcome. Do you mean the issue is `yarn link` doesn't create a `node_modules/.bin/ncc` entry? That seems like a linking issue to me, although the fix would be to just run `node_modules/ncc/dist/ncc/cli.js` instead.

PR #344, opened 2019-04-03
Fix watcher behaviour

guybedford:
This fixes the watcher regression in #342, and includes a full end-to-end test of the watching behaviour to avoid any further regressions in future.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -34,8 +34,9 @@ Options:
 let api = false;
 if (require.main === module) {
   runCmd(process.argv.slice(2), process.stdout, process.stderr)
-  .then(() => {
-    process.exit()
+  .then((watching) => {
+    if (!watching)
+      process.exit();
   })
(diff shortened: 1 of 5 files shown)


styfle:
@guybedford It looks like CI is failing

src/cli.js
```diff
@@ -315,6 +316,7 @@ async function runCmd (argv, stdout, stderr) {
           startTime = Date.now();
           stdout.write('File change, rebuilding...\n');
         });
+        return true;
```
styfle: Is this correct? You return a promise below but here you return a boolean?

guybedford: Yes, this is a result value to indicate we're going into watch mode.

guybedford:
Ok looking into it.

guybedford:
Ok the test should be working now... the test case was overriding the ncc self-build :P

PR #351, opened 2019-04-11
Expose custom watcher

guybedford:
This allows `watch` to be a custom Webpack watcher API.

The example API can be seen in https://github.com/zeit/ncc/compare/hook-watch?expand=1#diff-64e6d0a09767b69fdf92af31d738a902, where a sample watcher is implemented.

The test case performance only kicks in on the third run due to our timestamps being wrong (so taking two runs to sync) in the example case, but a real-world example the correct performance should work on the first run.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -224,6 +224,28 @@ module.exports = (
     .then(finalizeHandler);
   }
   else {
+    if (typeof watch === 'object') {
+      if (!watch.watch)
+        throw new Error('Watcher class must be a valid Webpack WatchFileSystem class instance (https://github.com/webpack/webpack/blob/master/lib/node/NodeWatchFileSystem.js)');
+      compiler.watchFileSystem = watch;
+      watch.inputFileSystem = compiler.inputFileSystem;
+    }
+    else {
+      // Used for inspecting Webpack watcher API
+      /*
+      const w = compiler.watchFileSystem.watch;
+      compiler.watchFileSystem.watch = function (files, dirs, missing, startTime, options, callback, callbackUndelayed) {
+        const cb = function () {
+          const timestamps = {};
+          for (const [key, value] of arguments[1].entries())
+            timestamps[key] = value;
+          console.log(JSON.stringify(timestamps));
+          callback.apply(this, arguments);
+        };
+        return w.call(this, files, dirs, missing, startTime, options, cb, callbackUndelayed);
+      };
+      */
+    }
     let cachedResult;
     watcher = compiler.watch({}, (err, stats) => {
       if (err)
@@ -232,7 +254,7 @@ module.exports = (
         return watchHandler({ err: stats.toString() });
       const returnValue = finalizeHandler();
       // clear output file system
-      mfs.data = {};
+      // mfs.data = {};
       if (watchHandler)
         watchHandler(returnValue);
       else
--- a/src/index.js
+++ b/src/index.js
@@ -230,22 +230,6 @@ module.exports = (
       compiler.watchFileSystem = watch;
       watch.inputFileSystem = compiler.inputFileSystem;
     }
-    else {
-      // Used for inspecting Webpack watcher API
-      /*
-      const w = compiler.watchFileSystem.watch;
-      compiler.watchFileSystem.watch = function (files, dirs, missing, startTime, options, callback, callbackUndelayed) {
-        const cb = function () {
-          const timestamps = {};
-          for (const [key, value] of arguments[1].entries())
-            timestamps[key] = value;
-          console.log(JSON.stringify(timestamps));
-          callback.apply(this, arguments);
-        };
-        return w.call(this, files, dirs, missing, startTime, options, cb, callbackUndelayed);
-      };
-      */
-    }
     let cachedResult;
     watcher = compiler.watch({}, (err, stats) => {
       if (err)
@@ -254,7 +238,7 @@ module.exports = (
         return watchHandler({ err: stats.toString() });
       const returnValue = finalizeHandler();
       // clear output file system
-      // mfs.data = {};
+      mfs.data = {};
       if (watchHandler)
         watchHandler(returnValue);
       else
--- a/src/index.js
+++ b/src/index.js
@@ -237,8 +237,6 @@ module.exports = (
       if (stats.hasErrors())
         return watchHandler({ err: stats.toString() });
       const returnValue = finalizeHandler();
-      // clear output file system
-      mfs.data = {};
       if (watchHandler)
         watchHandler(returnValue);
       else
--- a/test/watcher.test.js
+++ b/test/watcher.test.js
@@ -58,10 +58,13 @@ class CustomWatchFileSystem {
     this.dirs = new Set(dirs);
     this.missing = new Set(missing);
 
+    // empty object indicates "unknown" timestamp
+    // (that is, not cached)
     for (const item of files)
       this.timestamps.set(item, {});
     for (const item of dirs)
       this.timestamps.set(item, {});
+    // null represents "no file"
     for (const item of missing)
       this.timestamps.set(item, null);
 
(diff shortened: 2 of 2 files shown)


test/watcher.test.js
```diff
@@ -0,0 +1,131 @@
+const path = require('path');
+const ncc = (global.coverage || true) ? require("../src/index") : require("../");
```
styfle: What's the purpose of `global.coverage` here?

src/index.js
```diff
@@ -224,6 +224,28 @@ module.exports = (
...
   }
   else {
+    if (typeof watch === 'object') {
+      if (!watch.watch)
+        throw new Error('Watcher class must be a valid Webpack WatchFileSystem class instance (https://github.com/webpack/webpack/blob/master/lib/node/NodeWatchFileSystem.js)');
+      compiler.watchFileSystem = watch;
+      watch.inputFileSystem = compiler.inputFileSystem;
+    }
+    else {
+      // Used for inspecting Webpack watcher API
```
styfle: I don't understand this part. This else is for `typeof watch !== 'object'`? Is the assumption that `watch` is `true` at this point?

test/watcher.test.js
guybedford: This is the same thing we do in `test/index.test.js` which is that the coverage build uses the source files because the coverage maps only work against the original sources.

src/index.js
guybedford: Exactly - `watch: true` is the normal API, while `watch: new Watcher()` is the hooked API version.

guybedford:
Managed to fix this up, and it was only a few subtle things that were wrong it turns out.

The performance is now on par with the --watch API, but through a fully custom watcher hook.

The reason this wasn't working previously was a bit of a catch - if the underlying file hasn't actually changed on the file system, then the build files will not be output by webpack. But this should never happen in any real watching scenario.

test/watcher.test.js
```diff
@@ -0,0 +1,140 @@
...
+      }
+      else {
+        console.timeEnd('Watched Build');
+      }
+      if (buildCnt === 3) {
+        close();
+        fs.writeFileSync(buildFile, fs.readFileSync(buildFile).toString().slice(0, -2));
+      }
+    });
+    rebuild(() => {
```
guybedford: This hook runs on every rebuild start, by ncc.

test/watcher.test.js
```diff
@@ -0,0 +1,140 @@
...
+    }, function watchEnd () {
+      resolve();
+    });
+
+    console.time('First Build');
+    const { handler, rebuild, close } = ncc(buildFile, {
+      watch: watcher
+    });
+    
+    handler(({ err, code, map, assets, permissions }) => {
```
guybedford: This hook runs on every build completion, called by ncc.

test/watcher.test.js
```diff
@@ -0,0 +1,140 @@
...
+  }
+}
+
+jest.setTimeout(20000);
+
+it('Should support custom watch API', async () => {
+  let buildCnt = 0;
+  const buildFile = path.resolve('./test/integration/twilio.js');
+  await new Promise((resolve, reject) => {
+    const watcher = new CustomWatchFileSystem(function watchStart (files, dirs, missing) {
```
guybedford: This is a webpack WatchFileSystem class API, which we can pass to the ncc `watch` option.

test/watcher.test.js
```diff
@@ -0,0 +1,140 @@
...
+const ncc = global.coverage ? require("../src/index") : require("../");
+const fs = require('fs');
+
+// Based on the NodeWatchFileSystem class at:
+// - https://github.com/webpack/webpack/blob/master/lib/node/NodeWatchFileSystem.js
+// which in turn exposes:
+// - https://www.npmjs.com/package/watchpack
+
+class CustomWatchFileSystem {
+  constructor(watchStart, watchEnd) {
```
guybedford: `watchStart` and `watchEnd` here are convenience hooks in this test implementation.

test/watcher.test.js
```diff
@@ -0,0 +1,140 @@
...
+    this.files = undefined;
+    this.dirs = undefined;
+    this.missing = undefined;
+    this.timestamps = new Map();
+
+    // this will be populated for us by ncc
+    this.inputFileSystem = undefined;
+  }
+  
+  triggerChanges (changed, removed) {
```
guybedford: This function is a convenience hook in this test implementation, allowing us to "mock" a watch change event, where `changed` is an array of file paths changed, and `removed` is an array of file paths deleted. Webpack usually aggregates these over a time interval.

test/watcher.test.js
```diff
@@ -0,0 +1,140 @@
...
+          accuracy: 10,
+          timestamp: newTime
+        });
+      for (const file of removed)
+        this.timestamps.set(file, null);
+
+      for (const file of changed)
+        this.inputFileSystem.purge(file);
+      for (const file of removed)
+        this.inputFileSystem.purge(file);
```
guybedford: Note how the Watch API is responsible for puring the input file system.

test/watcher.test.js
```diff
@@ -0,0 +1,140 @@
...
+  
+  triggerChanges (changed, removed) {
+    if (!this.paused) {
+      const newTime = +Date.now();
+      for (const file of changed)
+        this.timestamps.set(file, {
+          safeTime: newTime + 10,
+          accuracy: 10,
+          timestamp: newTime
+        });
```
guybedford: Maintaining the file stamps from the watcher is integral to caching.

test/watcher.test.js
```diff
@@ -0,0 +1,140 @@
...
+    // null represents "no file"
+    for (const item of missing)
+      this.timestamps.set(item, null);
+
+    this.paused = false;
+    this.changeCallback = changeCallback;
+
+    // ...Start watching files, dirs, mising
+    setImmediate(() => {
+      this.watchStart(files, dirs, missing);
```
guybedford: This is where a real watch implementation would set the listeners.

test/watcher.test.js
```diff
@@ -0,0 +1,140 @@
...
+        null,
+        this.timestamps,
+        this.timestamps,
+        removed
+      );
+    }
+  }
+
+  // This is called on every rebuild
+  watch (files, dirs, missing, startTime, options, changeCallback) {
```
guybedford: This is the only part of the API that webpack calls and expects the correct return value for.

styfle:
I want @TooTallNate to take a look since this will be used for `now dev`

test/watcher.test.js
```diff
@@ -0,0 +1,140 @@
...
+    this.inputFileSystem = undefined;
+  }
+  
+  triggerChanges (changed, removed) {
+    if (!this.paused) {
+      const newTime = +Date.now();
+      for (const file of changed)
+        this.timestamps.set(file, {
+          safeTime: newTime + 10,
+          accuracy: 10,
```
TooTallNate: What do `safeTime` and `accuracy` represent?

test/watcher.test.js
```diff
@@ -0,0 +1,140 @@
...
+        });
+      for (const file of removed)
+        this.timestamps.set(file, null);
+
+      for (const file of changed)
+        this.inputFileSystem.purge(file);
+      for (const file of removed)
+        this.inputFileSystem.purge(file);
+
+      this.changeCallback(
```
TooTallNate: What are the arguments passed to `changeCallback()`?

test/watcher.test.js
guybedford: I've only been able to infer these myself looking at the implementation of http://npmjs.org/package/watchpack, perhaps @sokra can clarify further.

But I believe:

* safeTime represents an upper bound of the actual mtime, given that the mtime on its own might not be enough to trigger a cache invalidation
* timestamp being the mtime
* accuracy being the milliseconds from the best known timestamp that the safeTime is set to

test/watcher.test.js
guybedford: * The first null argument is the error
* The second two arguments are the same timestamp object. I believe the second one is a legacy Webpack API where the timestamps object was separate to another contextual timestamps object.
* removed is the array of file path strings which were deleted.

Issue #353, opened 2019-04-15
Incorrect source maps for typescript


styfle:
## Steps to reproduce

```
git clone https://github.com/styfle/ncc-bug-stack
cd ncc-bug-stack
git checkout ts
yarn
yarn build
node dist
```

Then visit http://localhost:3000

Notice the stack trace shows `ncc-bug-stack/index.ts:6:1`

## Expected

The expected line number is [index.ts:4](https://github.com/styfle/ncc-bug-stack/blob/ts/index.ts#L4), not `6`.

Also the column number is wrong (but that is less important).

guybedford:
I just ran the example build in the source map visualization tool (https://sokra.github.io/source-map-visualization/#custom) and the source map is definitely 100% correct (see below)
![Screenshot from 2019-04-18 19-11-40](https://user-images.githubusercontent.com/598730/56378748-3b920400-620e-11e9-899e-2dc69f63a992.png)

So this means the issue must lie with https://github.com/evanw/node-source-map-support or v8 itself here unfortunately.

guybedford:
I also tried this against the unbuilt non-ncc version of source-map-support and the same applies, so it's definitely the source map support package itself.

guybedford:
Ahh wait, in the example above the `index.ts` is not the original ts file but actually the ts converted to js....

So it should be as simple as fixing up the sourcemap chaining support here.

PR #355, opened 2019-04-18
Fix sourcemap chaining

guybedford:
This fixes the loaders to support sourcemap chaining properly, and includes a simple integration test to show the TypeScript sourcemaps are that the right offset.

Tracking Issue: https://github.com/zeit/ncc/issues/353

There will likely be more work to be done here in due course, which we can improve over time (in terms of ensuring the fidelity of the source maps, not their accuracy).

Prerequisite on https://github.com/zeit/webpack-asset-relocator-loader/pull/21.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -94,7 +94,7 @@ module.exports = (
       nodeEnv: false,
       minimize: false
     },
-    devtool: sourceMap ? "cheap-source-map" : false,
+    devtool: sourceMap ? "source-map" : false,
     mode: "production",
     target: "node",
     output: {
--- a/test/cli.js
+++ b/test/cli.js
@@ -29,5 +29,11 @@
     args: ["build", "-o", "tmp", "--watch", "test/fixtures/no-dep.js"],
     timeout: 500,
     expect: { timeout: true }
+  },
+  {
+    args: ["run", "test/fixtures/fail.ts"],
+    expect (code, stdout, stderr) {
+      return code === 1 && stderr.toString().indexOf('fail.ts:2:1') !== -1;
+    }
   }
 ]
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "@google-cloud/firestore": "^0.19.0",
     "@sentry/node": "^4.3.0",
     "@tensorflow/tfjs-node": "^0.3.0",
-    "@zeit/webpack-asset-relocator-loader": "0.3.7",
+    "@zeit/webpack-asset-relocator-loader": "0.4.0",
     "analytics-node": "^3.3.0",
     "apollo-server-express": "^2.2.2",
     "arg": "^4.1.0",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "@google-cloud/firestore": "^0.19.0",
     "@sentry/node": "^4.3.0",
     "@tensorflow/tfjs-node": "^0.3.0",
-    "@zeit/webpack-asset-relocator-loader": "0.4.0",
+    "@zeit/webpack-asset-relocator-loader": "0.4.1",
     "analytics-node": "^3.3.0",
     "apollo-server-express": "^2.2.2",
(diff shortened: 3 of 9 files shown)


guybedford:
We should also keep an eye on performance with this change.

test/cli.js
```diff
@@ -29,5 +29,11 @@
     args: ["build", "-o", "tmp", "--watch", "test/fixtures/no-dep.js"],
     timeout: 500,
     expect: { timeout: true }
+  },
+  {
+    args: ["run", "test/fixtures/fail.ts"],
```
styfle: This file doesn't exist 🤔

src/index.js
```diff
@@ -94,7 +94,7 @@ module.exports = (
       nodeEnv: false,
       minimize: false
     },
-    devtool: sourceMap ? "cheap-source-map" : false,
+    devtool: sourceMap ? "source-map" : false,
```
styfle: What's the difference between `cheap-source-map` and `source-map`?

test/cli.js
guybedford: Thanks, added.

src/index.js
guybedford: I have absolutely no idea. I just thought it best not to go for the cheap option :P

guybedford: Looking at https://webpack.js.org/configuration/devtool/ I believe this is actually a fidelity thing :)

We could likely go back to cheap-source-map for performance.

guybedford: I've pushed a commit setting this back to cheap-source-map. Feel free to revert though if you feel otherwise, but perhaps perf should be a priority?

styfle: @guybedford Can you measure the perf difference? If source-map is 80% slower than cheap-source-map then we might want to stick with cheap. But if source-map is 10% slower than cheap-source-map then we might want to use `source-map` for the fidelity.

guybedford: Just did a very quick test, and there were no significant differences between the two at all.

test/cli.js
```diff
@@ -29,5 +29,11 @@
...
+    expect (code, stdout, stderr) {
+      return code === 1 && stderr.toString().indexOf('fail.ts:2:1') !== -1;
```
styfle: So the column number is still wrong (1), even after changing `devtool`?

styfle:
@guybedford I just checked out this branch and built to the code to see how it would work with #353 and it is actually worse.

```
git clone https://github.com/styfle/ncc-bug-stack
cd ncc-bug-stack
git checkout ts
yarn
rm -rf dist && ../ncc/dist/ncc/cli.js build index.ts --source-map
node dist
```

Visit http://localhost:3000

Expected: `ncc-bug-stack/index.ts:4`

Actual: `ncc-bug-stack/dist/index.js:61`

guybedford:
@styfle interesting, must have changed since my first tests. I've pushed up a patch with a fix.

styfle:
First I ran this in the `ncc` directory

```
git pull && rm -rf node_modules && yarn && yarn build
```

Then I ran this in the `ncc-bug-stack` directory

```
rm -rf dist && ~/Code/zeit/ncc/dist/ncc/cli.js cache clean && ~/Code/zeit/ncc/dist/ncc/cli.js build index.ts --source-map && node dist
```

I still see the same (incorrect) error:

```
Error: Should throw
    at Server.handler (/Users/styfle/Code/foo/ncc-bug-stack/dist/index.js:61:19)
    at Server.emit (events.js:189:13)
    at parserOnIncoming (_http_server.js:676:12)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:109:17)
```

I even tried disabling chrome caching 🤷‍♂️

styfle:
Ok so change `tsconfig.json` to enable `sourceMap:true` fixed the problem.

We'll have to make sure that user's enable this feature in their TS projects.

PR #375, opened 2019-05-06
Only output build errors, not warnings, on error

guybedford:
This resolves https://github.com/zeit/ncc/issues/373, so that TypeScript errors will only display the direct error message, instead of all the possible Webpack warnings first.


src/index.js
```diff
@@ -133,7 +133,8 @@ module.exports = (
         ))
           return callback();
       } while (pkgPath.length > (pkgPath = pkgPath.substr(0, pkgPath.lastIndexOf(sep, pkgPath.length - 15 - request.length)) + sep + 'node_modules' + sep + request).length);
-      console.error(`ncc: Module directory "${context}" attempted to require "${request}" but could not be resolved, assuming external.`);
+      if (debugLog)
+        console.error(`ncc: Module directory "${context}" attempted to require "${request}" but could not be resolved, assuming external.`);
```
styfle: I'm torn on this one. In production builds, its nice to know what was assumed as external. On the other hand, I don't think I have seen and bugs caused by the assumption of `external`

test/cli.js
```diff
@@ -31,9 +31,15 @@
     expect: { timeout: true }
   },
   {
-    args: ["run", "test/fixtures/fail.ts"],
+    args: ["run", "test/fixtures/ts-error1/fail.ts"],
```
styfle: Why was this changed?

src/index.js
guybedford: Ok, if we're not sure lets leave it in for now.

test/cli.js
guybedford: Because ts-loader will throw errors for all ts files in the project, such that both error messages become the error message for the one file otherwise!

I kind of see this as something we should address in https://github.com/zeit/ncc/issues/362.

src/index.js
```diff
@@ -217,8 +217,13 @@ module.exports = (
         if (err) return reject(err);
         compiler.close(err => {
           if (err) return reject(err);
-          if (stats.hasErrors())
-            return reject(new Error(stats.toString()));
+          if (stats.hasErrors()) {
+            let errLog = '';
+            stats.compilation.errors.forEach((err, index) => {
```
styfle: This is an interesting way to do it. I would have done

```
stats.compilation.errors.map(err => err.message).join('\n')
```

But I guess yours only loops once?

guybedford: Sure, let's keep it simple.

styfle:
@guybedford CI is failing though

guybedford:
Just figuring out why CI is failing... should be able to get this green shortly.

guybedford:
Ok, should pass this run with some luck.

PR #378, opened 2019-05-09
New analysis updates

guybedford:
This updates to webpack-asset-relocator-loader@0.5.0-beta.3, which includes resolutions for the following ncc bugs:

* https://github.com/zeit/ncc/issues/290 (integration test included)
* https://github.com/zeit/ncc/issues/376 (integration test included)
* https://github.com/zeit/ncc/issues/216
* https://github.com/zeit/ncc/issues/349 (integration test included)
* https://github.com/zeit/ncc/issues/41
* https://github.com/zeit/ncc/issues/216
* https://github.com/zeit/ncc/issues/297
* https://github.com/zeit/ncc/issues/148
* https://github.com/zeit/ncc/issues/156
* https://github.com/zeit/ncc/issues/248
* https://github.com/zeit/ncc/issues/383

The full description of the analysis changes is in https://github.com/zeit/webpack-asset-relocator-loader/pull/30.

In ncc itself we also change not found errors into runtime errors and remove the external warnings for not-found requires.


guybedford:
In addition - package resolutions are attempted for resolution at runtime before throwing a not found error.

styfle:
> This change also allows us to remove the entire code path around the custom externals function we have, which is currently double-resolving all package resolutions (and incurring the statting and performance costs of that). I haven't removed it here because the externals warning is a feature we need to consider if we want to remove, but I would personally be quite happy removing those logs too.

If these warnings are decreasing performance, we should remove them.

Like you mentioned before: if the module require is tried at runtime and not found, then there will be an error instead. I think this error is pretty self-explanatory.

```
Error: Cannot find module 'unknown'
```

styfle:
Can all those issues be closed now?

guybedford:
Yes, all those issues can be closed!

Issue #315, opened 2019-03-18
Change `amd` option to false


styfle:
Since `ncc` is a node compiler and node can't use AMD, we can disable it.

There is a new option that will be released in the next version of webpack `amd: false`

https://github.com/webpack/webpack/pull/8890

guybedford:
We already have this - https://github.com/zeit/ncc/blob/master/src/index.js#L170.

styfle:
@guybedford I don't believe that is the same setting because this feature wasn't introduced until a couple weeks ago in [webpack@4.30.0](https://github.com/webpack/webpack/releases/tag/v4.30.0).

The line you linked to is for the parse option but this is a top-level webpack option.

guybedford:
Ahh I see. Do we need a separate PR to bring it to the Webpack 5 branch (used by ncc) as well then?

styfle:
@guybedford I searched the release notes for webpack 5 and don't see it announced so I think the answer is yes, we need a separate PR for webpack 5

guybedford:
Created the port PR at https://github.com/webpack/webpack/pull/9074.

styfle:
Fixed in 388

PR #388, opened 2019-05-15
Webpack deterministic module ids

guybedford:
This adds the Webpack options for deterministic module ids.

This also includes the Webpack 5 alpha.12 upgrade.

Also [resolves]: https://github.com/zeit/ncc/issues/315

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -91,6 +91,7 @@ module.exports = (
       name: `ncc_${hashOf(entry)}`,
       version: nccVersion
     },
+    amd: false,
     optimization: {
       nodeEnv: false,
       minimize: false,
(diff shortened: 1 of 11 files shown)


styfle:
Does this version of webpack include the feature to implement #315?

guybedford:
Ah good point - that was just released in alpha.12 as well so I've added that to this PR too.

src/index.js
```diff
@@ -91,9 +91,13 @@ module.exports = (
       name: `ncc_${hashOf(entry)}`,
       version: nccVersion
     },
+    amd: false,
     optimization: {
       nodeEnv: false,
-      minimize: false
+      minimize: false,
+      moduleIds: 'deterministic',
```
sokra: deterministic module and chunk ids is the default in production mode anyway.

PR #401, opened 2019-05-22
Change source code to typescript

### LINKED ISSUE #343: Change source code to TypeScript
We like TypeScript. Let's write the source code for ncc in TypeScript.

Note: default exports are painful to deal with CJS/ESM interop so maybe use a named export instead.

styfle:
Work in progress...

Fixes #343


styfle:
@guybedford This was a little bit more work than I anticipated.

I found a few unused imports and variables that I removed, but some of them I'm not sure if they are safe to remove.

Can you take a look at the existing errors from `yarn build` to see how we can resolve the ts errors?

styfle:
@guybedford This might be a good opportunity to change the API (not the CLI) to something like the following:

```ts
import { build, watch } from '@zeit/ncc';

const { code, map, assets } = await build(entry, options);

const { handler, rebuild, close } = await watch(entry, options);
```

Thoughts?

guybedford:
This looks great to me. The CI seems to be failing because `yarn test-coverage` runs a test against the original sources in the `src` folder. It should be possible to update this to use the tsc compiled sources I think.

As for changing the watcher API to be a specific call - that seems like a great suggestion. My only concern with named exports is that it's worth remembering that this is what it looks like in `node --experimental-modules`:

```js
import ncc from 'ncc';
const { build, watch } = ncc;
```

so I'm not sure we're quite there on the user experience around named exports, but I agree with the overall API idea if you want to do that.

futagoza:
@guybedford  If you use both named exports and a default export it should be fine, shouldn't it?

```ts
// @zeit/ncc

export function build() { }

export function watch() { }

export default { build, watch }

// named importing

import { build, watch } from '@zeit/ncc';

// default importing

import ncc from '@zeit/ncc';

```

PR #403, opened 2019-05-22
Disable v8cache in ncc build

guybedford:
This resolves https://github.com/zeit/ncc/issues/395 in disabling the v8cache in the ncc self-build.

The underlying issue seems to be that there was a report of a possible v8 cache bug on a specific architecture, so that is enough of an indication of instability for us to rather forgo this minor performance benefit.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -59,7 +59,7 @@ async function main() {
   // detect unexpected asset emissions from core build
   const unknownAssets = [
     ...Object.keys(cliAssets),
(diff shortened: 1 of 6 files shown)


guybedford:
It turns out the performance drop is actually quite significant as can be seen from the fact that CI is now timing out :(

guybedford:
Can we extend the CI timeout beyond 10 mins?

guybedford:
Agreed, closing for now.

PR #404, opened 2019-05-22
Disable v8cache on Windows only

guybedford:
This is an alternative workaround for https://github.com/zeit/ncc/issues/395 given that disabling the v8 cache in https://github.com/zeit/ncc/pull/403 turned out to be such a large performance regression.

Since the bug report came from Azure on Node.js 10, this just disables using the v8cache in Windows only as part of the cache runtime itself.

### DIFF
--- a/src/sourcemap-register.js
+++ b/src/sourcemap-register.js
@@ -1,3 +1,0 @@
-// note: this file is overriden by `scripts/build`
-// and substituted for the production release
-module.exports = require("source-map-support/register");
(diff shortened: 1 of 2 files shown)


src/sourcemap-register.js
```diff
@@ -1,3 +0,0 @@
-// note: this file is overriden by `scripts/build`
-// and substituted for the production release
-module.exports = require("source-map-support/register");
```
styfle: Did you intend to remove this?

guybedford: Yes it's actually not used anymore.

PR #416, opened 2019-06-11
Add google cloud firestore integration test

guybedford:
This adds an integration test for `require('@google-cloud/firestore')`, as in https://github.com/zeit/ncc/issues/412.


styfle:
@guybedford So it looks like it passes without any changes. Do we just need to do a new release?

PR #429, opened 2019-06-17
Disable separators in entries

guybedford:
Currently there are a number of relative pathing bugs when running builds with entry names containing separators. Until these are resolved, this provides a clear error message that these cases aren't supported currently.


guybedford:
Closing to rather revert the multi-entry build support.

PR #430, opened 2019-06-18
Revert "Support multiple entry inputs (#410)"

guybedford:
This reverts the support for multiple entry points in ncc.

The hope was that this feature would allow us to share module instances between multiple entry points to support these builds (eg Next.js has multiple entry points that share common modules that would benefit from this, as well as ncc itself, in reducing the package size, execution footprint, and risk of bugs with shared state).

Unfortunately after hitting some bugs and discussing further with @sokra it turns out that Webpack simply doesn't support this kind of a chunking model. So it would be best for us not to offer this feature at all.

We can then go back to our former API from 0.18, and post a new 0.20 with this approach.


styfle:
@guybedford This is a complete git revert, correct? You didn't manually make these changes?

guybedford:
I had to handle some conflicts, but yes it's a git revert.

PR #431, opened 2019-06-18
Update to Webpack 5.0.0-alpha.16

guybedford:
This updates Webpack to the latest alpha release.


src/index.js
```diff
@@ -331,7 +331,7 @@ module.exports = (
     }
 
     for (const chunk of stats.compilation.chunks) {
-      finalizeOutput(chunk.files[0]);
+      finalizeOutput(chunk.files.values().next().value);
```
styfle: There's a merge conflict here after reverting the API in the other PR.

In fact, there is no `finalizeOutput()` function anymore 🤷‍♂

guybedford: Just rebased.

test/integration/twilio.js
```diff
@@ -7,3 +7,4 @@ try {
   }
 }
 
+
```
styfle: The only code change is a newline? I thought you had to change the chunkFiles?

guybedford: Nicely spotted, yes that shouldn't be there.

PR #445, opened 2019-07-02
ncc internal typescript build bug workaround

guybedford:
This provides a workaround for https://github.com/zeit/ncc/issues/435 by using the previous TypeScript build.

Hopefully we can get a proper bug fix in soon as well to avoid this workaround.


styfle:
Seems like the tests didn't like it

styfle:
Looks good 👍

PR #446, opened 2019-07-02
Treat uglify-es like uglify-js

guybedford:
Due to the same awful code pattern we can't analyze that makes it unbuildable by Webpack, this extends the `uglify-js` treatment to `uglify-es`.

### DIFF
--- a/src/loaders/empty-loader.js
+++ b/src/loaders/empty-loader.js
@@ -12,7 +12,7 @@ function getPackageBase(id) {
   }
 }
 
-const emptyModules = { 'uglify-js': true };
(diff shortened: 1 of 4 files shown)


src/loaders/empty-loader.js
```diff
@@ -12,7 +12,7 @@ function getPackageBase(id) {
   }
 }
 
-const emptyModules = { 'uglify-js': true };
+const emptyModules = { 'uglify-js': true, 'uglify-es': true };
```
styfle: Can you add a test for this one?

guybedford: Sure.

Issue #459, opened 2019-07-15
Expose Webpack resolve.alias & resolve.modules


shcca:
It'd be helpful if Webpack resolve [alias](https://webpack.js.org/configuration/resolve/#resolvealias) and [modules](https://webpack.js.org/configuration/resolve/#resolvemodules) were exposed for programmatic use.

In my particular case, I'm using @zeit/ncc to bundle AWS Lambdas, but a problem arises with AWS Lambda Layers because Layers exist outside of Lambda dirs. Thus, Layer modules don't resolve.

For example:

`lambda/index.js`
```javascript
'use strict';

const layerModule1 = require('layer-module1');
const layerModule2 = require('opts/layer-module2');

// rest of lambda
```

`layer-module1` and `layer-module2` won't resolve during bundling.

But with programmatic @zeit/ncc and exposed Webpack resolve alias and modules, there could be something like:
```javascript
require('@zeit/ncc')('/path/to/input', {
  resolveAlias: {
    opts: path.resolve('../layer/nodejs/opts')
  },
  resolveModules: [path.resolve('../layer/nodejs/node_modules')]
}).then(({ code, map, assets }) => {
  .....
})
```

Beyond my specific example, I think others looking for a no-config Node.js bundle option would benefit from exposing alias and modules. It doesn't add overhead and doesn't cross the line of doing too much (ie, just use Webpack).

I'm going to provide a PR outline. Thanks for your time and consideration.

PR #460, opened 2019-07-15
Expose Webpack resolve alias and modules properties

shcca:
This is in regard to #459

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -83,7 +83,11 @@ require('@zeit/ncc')('/path/to/input', {
   watch: false, // default
   v8cache: false, // default
   quiet: false, // default
-  debugLog = false // default
+  debugLog = false, // default
+  // https://webpack.js.org/configuration/resolve/#resolvealias
+  resolveAlias: {} // default,
+  // https://webpack.js.org/configuration/resolve/#resolvemodules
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -83,7 +83,11 @@ require('@zeit/ncc')('/path/to/input', {
   watch: false, // default
   v8cache: false, // default
   quiet: false, // default
-  debugLog = false // default
+  debugLog = false, // default
+  // https://webpack.js.org/configuration/resolve/#resolvealias
+  resolveAlias: {}, // default
+  // https://webpack.js.org/configuration/resolve/#resolvemodules
+  resolveModules: [] // default
```
styfle: I think we want to avoid leaking webpack details through to the ncc API in case we need to swap webpack for rollup or some other alternative.

fregante: Other tools likely have similar options; Rollup for sure does.

Changing build tool would definitely be a breaking change, if it ever happens, so I don't particularly see the issue with supporting a particular flavor of options.

This option would reduce my bundle by dropping 2 unused heavy sub-dependencies. I just realized that my built file is 1.2MB due a couple of dependencies I don't really use.

PR #470, opened 2019-08-31
Fix issue with TypeScript paths with `allowJs`

### LINKED ISSUE #469: TypeScript with allowJS and absolute imports fails at runtime
I've created a repo to reproduce this issue:
https://github.com/novascreen/ncc-ts-allowjs

TypeScript is configured to `allowJS`, which we need in a project we are transitioning over to TS gradually. We also have `baseUrl`/`paths` configured to be able to do absolute imports of modules in `src`.
The ncc build runs fine and absolute imports of TS files work perfectly, but when running the output we get a `Cannot find module` for a JS file we tried to import using an absolute import.

In summary:
- relative TS import: ✔
- absolute TS import: ✔
- relative JS import: ✔
- absolute JS import: ❌

I would expect this to either work or hope to get at least an error during build. We can work around this using relative imports for the JS files, but it's easy to accidentally forget and then get a runtime error once the app is built and deployed 😬

novascreen:
Closes #469

I noticed something when I added the test, when you run the existing `output.js` of `tsconfig-paths` you also get "Cannot find module", not sure if the output there is intended to be different from `output-coverage.js`, but thought i'd point it out.

The `tsconfig-paths-allowjs/output-coverage.js` shows the difference between adding `js` to the extensions of `TsconfigPathsPlugin` and not having it.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -71,7 +71,7 @@ module.exports = (
   // we need to catch here because the plugin will
   // error if there's no tsconfig in the working directory
   try {
-    resolvePlugins.push(new TsconfigPathsPlugin({ silent: true, extensions: [".ts", ".tsx", ".js"] }));
+    resolvePlugins.push(new TsconfigPathsPlugin({ silent: true, extensions: SUPPORTED_EXTENSIONS }));
 
     const tsconfig = tsconfigPaths.loadConfig();
     if (tsconfig.resultType === "success") {
--- a/src/index.js
+++ b/src/index.js
@@ -71,7 +71,7 @@ module.exports = (
   // we need to catch here because the plugin will
   // error if there's no tsconfig in the working directory
   try {
-    resolvePlugins.push(new TsconfigPathsPlugin({ silent: true }));
+    resolvePlugins.push(new TsconfigPathsPlugin({ silent: true, extensions: [".ts", ".tsx", ".js"] }));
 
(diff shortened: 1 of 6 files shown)


novascreen:
Another thing, I wasn't sure whether i should only add `js` or make `extensions` be `SUPPORTED_EXTENSIONS`, let me know if i should change that.

src/index.js
```diff
@@ -71,7 +71,7 @@ module.exports = (
   // we need to catch here because the plugin will
   // error if there's no tsconfig in the working directory
   try {
-    resolvePlugins.push(new TsconfigPathsPlugin({ silent: true }));
+    resolvePlugins.push(new TsconfigPathsPlugin({ silent: true, extensions: [".ts", ".tsx", ".js"] }));
```
```suggestion
    resolvePlugins.push(new TsconfigPathsPlugin({ silent: true, extensions: SUPPORTED_EXTENSIONS }));
```

The docs mention that these should be the same.

Perhaps this should look for `allowJs` option before assigning the extensions.

> If you're using allowJs in tsconfig.json, or allow other non-TS extensions in webpack, make sure you set extensions option in sync with your webpack config.

novascreen: @styfle I believe I addressed both your suggestions, I did it in two steps, let me know if you want to keep the check against the tsconfig.json

PR #490, opened 2019-12-04
Add API alias options

legendecas:
Exposing `alias` options allowing to map one module to another and bundle the mapped module into the final archive.

Rollup alternatives: https://github.com/rollup/plugins/tree/master/packages/alias

Related: https://github.com/zeit/ncc/pull/460

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -36,6 +36,7 @@ module.exports = (
   {
     cache,
     externals = [],
+    alias = [],
     filename = "index.js",
     minify = false,
     sourceMap = false,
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -107,6 +108,12 @@ module.exports = (
   else if (typeof externals === 'object')
     Object.keys(externals).forEach(external => externalMap.set(external, externals[external]));
 
+  const aliasMap = {};
+  if (Array.isArray(aliasMap))
+    alias.forEach(alias => aliasMap[alias[0]] = alias[1]);
+  else if (typeof alias === 'object')
+    Object.keys(alias).forEach(from => aliasMap[from] = alias[from]);
```
```suggestion
    aliasMap = alias;
```

I think you can avoid iteration here an assign instead

morgs32:
I'm interested in this - wondering if it would enable transpiling other packages in a yarn workspace/monorepo? For example:

```
// packages/package1.js

import utils form 'package2'

```

If the `main` in package2's `package.json` is es6 or typescript, could it still be transpiled if we alias'd to it with a relative path instead of referencing the same package via `node_modules` in a yarn workspace (which points to the directory via symlink)

PR #497, opened 2020-01-03
Add transpileOnly support for TS projects if needed

arunoda:
This allows us to build faster while we are developing.
(As we are not much interest about the type accuracy)

With some tests, we can gain upto 2X time reduction when building.
I assume the benefit will be more than this for massive projects.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -45,7 +45,8 @@ module.exports = (
     v8cache = false,
     filterAssetBase = process.cwd(),
     quiet = false,
-    debugLog = false
+    debugLog = false,
+    transpileOnly = false
   } = {}
 ) => {
   if (!quiet) {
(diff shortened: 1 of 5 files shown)


ctrlplusb:
Amaze!

src/index.js
```diff
@@ -177,6 +178,7 @@ module.exports = (
           {
             loader: eval('__dirname + "/loaders/ts-loader.js"'),
             options: {
+              transpileOnly,
```
styfle: Looks good, we just need a test for this new behavior

arunoda: @styfle sure thing. Done.

arunoda:
Thanks @styfle.

> It looks like Circle CI is broken for macOS so we can probably ignore that for now. Ironically, GH Actions has the opposite problem where macOS passes and Linux fails 🤷‍♂ (see #494)

Then use Circle CI for linux and GA for MacOS.

PR #504, opened 2020-02-10
Fix minified source-map outputs format

legendecas:
Also added test infrastructure to validate generated source-maps.


src/index.js
```diff
@@ -366,7 +366,7 @@ module.exports = (
       // For some reason, auth0 returns "undefined"!
       // custom terser phase used over Webpack integration for this reason
       if (result.code !== undefined)
-        ({ code, map } = { code: result.code, map: result.map });
+        ({ code, map } = { code: result.code, map: map ? JSON.parse(result.map) : undefined });
```
```suggestion
        ({ code, map } = { code: result.code, map: sourceMap ? JSON.parse(result.map) : undefined });
```

Should this use the `sourceMap` flag instead? It looks like `map` has several meanings.

cc @guybedford

PR #510, opened 2020-02-20
Support `.js` imports from typescript

guybedford:
This resolves https://github.com/zeit/ncc/issues/508 in allowing the pattern for TypeScript code using `.js` extensions to reference other TypeScript files.

This is currently recommended by the TypeScript team (https://github.com/microsoft/TypeScript/issues/16577#issuecomment-309169829) as an approach for writing TypeScript that will include the file extensions in the output files when using `tsc` manually, resulting in Node.js and browser compatibility for the output.

With this PR, ncc doesn't have to break enabling this workflow for users, while remaining backwards compatible with the previous behaviour so it would be fine to release this as a minor.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -95,6 +95,13 @@ module.exports = (
           if (!err.missing || !err.missing.length)
             return callback(err);
           // make not found errors runtime errors
+          if (request.endsWith('.js'))
+            return resolve.call(resolver, context, path, request.slice(0, -3), resolveContext, function (err, result) {
+              if (!err) return callback(null, result);
+              if (!err.missing || !err.missing.length)
+                return callback(err);
+              callback(null, __dirname + '/@@notfound.js' + '?' + (externalMap.get(request) || request));
+            });
           callback(null, __dirname + '/@@notfound.js' + '?' + (externalMap.get(request) || request));
         });
       };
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -95,6 +95,13 @@ module.exports = (
           if (!err.missing || !err.missing.length)
             return callback(err);
           // make not found errors runtime errors
```
styfle: I think this comment "make not found errors runtime errors" is in the wrong place

styfle: And it might be worth adding a comment in the request.endsWith('.js') to explain why this is necessary.

styfle:
@guybedford Do we also need too add this behavior to [node-file-trace](https://github.com/zeit/node-file-trace)?

src/index.js
guybedford: This feature still applies separately - since if the path isn't found, we always ensure it is a runtime check.

This has been branched now as for backwards compat I wanted to first check the webpack resolver against `./file.js` before trying `./file` so it's effectively doing two resolutions checks on the fallback.

That said, I haven't guarded this by ensuring the parent is definitely a tsx? module.... let me add that.

guybedford:
@styfle thanks for the quick review. Yes ideally `node-file-trace` would include the same (if it doesn't support it already), and this is something I'd recommend checking in other workflows too.

src/index.js
guybedford: Fixed the comments.

guybedford:
Ok, I've included a guard that this extension fallback only happens when importing from another TypeScript file, and I also added a test here for nested imports of .ts.

PR #511, opened 2020-02-20
Add support for `.cjs` outputs

guybedford:
This PR adds support for `.cjs` output files from ncc, when the input file is using the `.cjs` extension.

As included in the readme comment, this is useful for packages setting `"type": "module"` in the package.json that want '.js' files in the package to be able to be modules, and thus need the '.cjs' extension on all CommonJS files.


readme.md
```diff
@@ -40,6 +40,9 @@ $ ncc build input.js -o dist
 
 Outputs the Node.js compact build of `input.js` into `dist/index.js`.
 
+> Note: If the input file is using a `.cjs` extension, then so will the corresponding output file.
+> This is useful for packages that want to use `.js` files as modules in native Node.js using
+> a `"type": "module"` in the package.json file.
```
styfle: What about an input file using the `.mjs` extension. Is that preserved?

src/cli.js
```diff
@@ -215,6 +215,7 @@ async function runCmd (argv, stdout, stderr) {
       let startTime = Date.now();
       let ps;
       const buildFile = eval("require.resolve")(resolve(args._[1] || "."));
+      const ext = buildFile.endsWith('.cjs') ? '.cjs' : '.js';
```
styfle: Let's create a function so this works with `.mjs` too.

```js
function getOutputExtension(input: string) {
  const ext = path.extname(input);
  if (ext === '.mjs' || ext === '.cjs') {
    return ext;
  }
  return '.js';
}
```

Then it can be reused in both places.

readme.md
guybedford: Supporting `.mjs` would actually be a separate feature, because it would involve changing the module format of the output file. Let's continue to track this in https://github.com/zeit/ncc/issues/512.

src/cli.js
guybedford: As mentioned above - yes this can easily extend to `.mjs`, but requires the support for an ES module output first, which should probably be a separate flag in its own PR.

src/cli.js
```diff
@@ -241,16 +242,16 @@ async function runCmd (argv, stdout, stderr) {
 
         outDir = outDir || resolve("dist");
         mkdirp.sync(outDir);
-        // remove all existing ".js" files in the out directory
+        // remove all existing ".js", ".mjs" and ".cjs" files in the out directory
         await Promise.all(
           (await new Promise((resolve, reject) =>
-            glob(outDir + '/**/*.js', (err, files) => err ? reject(err) : resolve(files))
+            glob(outDir + '/**/*.(js|mjs|cjs)', (err, files) => err ? reject(err) : resolve(files))
```
styfle: Should `mjs` be removed here?

guybedford: Sure, until we add support for mjs that makes sense. Added.

src/cli.js
```diff
@@ -241,16 +242,16 @@ async function runCmd (argv, stdout, stderr) {
 
         outDir = outDir || resolve("dist");
         mkdirp.sync(outDir);
-        // remove all existing ".js" files in the out directory
+        // remove all existing ".js", ".mjs" and ".cjs" files in the out directory
```
styfle:
```suggestion
        // remove all existing ".js" and ".cjs" files in the out directory
```

PR #518, opened 2020-03-16
Export build time not found packages on runtime

legendecas:
Exporting eval results.

### DIFF
--- a/src/@@notfound.js
+++ b/src/@@notfound.js
@@ -1,1 +1,1 @@
-__non_webpack_require__('UNKNOWN');
+module.exports = __non_webpack_require__('UNKNOWN');
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@ dist/**/*.js
 !test/integration/*.json
 !test/integration/*.js
 !test/integration/*.ts
+!test/integration/node-path
+!test/integration/node-path/*.js
 !test/unit
 !test/unit/**
 !dist/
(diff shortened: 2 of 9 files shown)


src/@@notfound.js
```diff
@@ -1 +1 @@
-__non_webpack_require__('UNKNOWN');
+module.exports = __non_webpack_require__('UNKNOWN');
```
styfle: What's the purpose of this change?

legendecas: There are cases that the packages are missing at ncc build time, yet presented at global scope or runtime. Without this fix, the `eval("require")('dep')` can succeed at runtime but the result is not exported.

styfle: > There are cases that the packages are missing at ncc build time, yet presented at global scope or runtime. 

Which cases? Can you add a test that fails but succeeds with this PR?

legendecas: Let's say we have a project referencing an optional dependency using

```js
try {
  foo = require('foo');
} catch {}
```

Package `foo` is not installed on build time since it's totally optional and should not be bundled into the products. So package `foo` will be transformed by loader `runtime-notfound` to `eval("require")("foo")` without assigning return value to `module.exports`. Since then the `require('foo')` can succeed without any error at runtime but the return value is `undefined`, not the dynamically required package module.

The test case `test/unit/runtime-notfound/output.js` is already presented in the repo, and is identical to the case above.

styfle: Sorry I should have been more clear. Can you add an **integration** test that fails with the latest ncc but succeeds with this PR?

legendecas: Integration test added, PTAL :)

PR #525, opened 2020-03-27
Add Node.js 12 CI

guybedford:
Update dependencies and add Node.js 12 to CI matrix.

### DIFF
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -2,8 +2,6 @@ const fs = require("fs");
 const { fork } = require("child_process");
 const ncc = global.coverage ? require("../src/index") : require("../");
 
-const isNode12 = process.version.startsWith('v12');
-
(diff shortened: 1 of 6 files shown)


test/index.test.js
```diff
@@ -123,6 +123,10 @@ for (const integrationTest of fs.readdirSync(__dirname + "/integration")) {
     }
   }
 
+  // Fixes todo
+  if (integrationTest === 'sharp.js' || integrationTest === 'oracledb.js' || integrationTest === 'leveldown.js')
```
styfle:
```suggestion
  const isNode12 = process.version.startsWith('v12');
  const skipTests = new Set(['sharp.js', 'oracledb.js', 'leveldown.js']);
  if (isNode12 && skipTests.has(integrationTest))
```

guybedford: Added.

PR #528, opened 2020-04-08
Properly populate cached js filename and fix misaligned columns

legendecas:
Previously, the cached script doesn't properly initiate with it's correlated JS filename. The filename in stack traces will be `evalmachine.<anonymous>` as a result, e.g.

```
Error: foo
    at main (evalmachine.<anonymous>:91:15)
    at Object.423 (evalmachine.<anonymous>:92:3)
    at __webpack_require__ (evalmachine.<anonymous>:22:30)
    at startup (evalmachine.<anonymous>:37:19)
    at evalmachine.<anonymous>:41:18
    at evalmachine.<anonymous>:44:10
    at Object.<anonymous> (/workspace/out/index.js:5:28)
    at Module._compile (internal/modules/cjs/loader.js:1147:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1167:10)
    at Module.load (internal/modules/cjs/loader.js:996:32)
```

With this patch, the stack traces can be properly mapped to its correlated JS file:

```
Error: foo
    at main (/workspace/out/index.js.cache.js:91:15)
    at Object.717 (/workspace/out/index.js.cache.js:92:3)
    at __webpack_require__ (/workspace/index.js.cache.js:22:30)
    at startup (/workspace/out/index.js.cache.js:37:19)
    at /workspace/out/index.js.cache.js:41:18
    at /workspace/out/index.js.cache.js:44:10
    at Object.<anonymous> (/workspace/out/index.js:7:28)
    at Module._compile (internal/modules/cjs/loader.js:1147:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1167:10)
    at Module.load (internal/modules/cjs/loader.js:996:32)
```


test/integration/stack-trace.js
```diff
@@ -0,0 +1,29 @@
...
+        process.exit(1);
+        return;
+      }
+      assert(stdout.match(/evalmachine.<anonymous>/) == null);
+      assert(stdout.match(__filename + '.cache.js') != null);
+    });
+    return;
+  }
+
+  console.log(new Error('foo'));
```
styfle, line 28: Did you intend to throw for both parent and child process?

legendecas: No, only on the child process. And no throwing is required, since we only like to check on the stack traces. So here we just log the error out and exit properly.

Issue #569, opened 2020-08-03
NCC does not pack licenses that are required to be included with their software


thboop:
We are hoping to use NCC to pack a javascript project and it's `node_modules` into a single js file we can execute. However, when packing the `node_modules`, certain licenses require that they be included with the software, for example MIT states:
> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

Could we also package the licenses for each of these node modules in order to stay compliant with these licenses?

For example, webpack has: https://www.npmjs.com/package/license-webpack-plugin

styfle:
Thanks for the suggestion!

Would you like to submit a PR to add this feature? Probably enable the plugin with the flag `--license`.

bryancross:
@styfle This is going to be necessary for redistributing anything built with `ncc`....

thboop:
Hey @styfle I've filed #570 , please let me know if there's anything else I can do to help!

Thanks so much!

PR #570, opened 2020-08-05
Add license flag to create license file when running ncc

thboop:
This PR addresses #569 

It adds the --license flag to use [license-webpack-plugin](https://www.npmjs.com/package/license-webpack-plugin) to generate a license file alongside the ncc output to adhere to the terms of common software licenses.

Currently, there is an open issue where that plugin outputs some warnings, so the output for a run looks like

```
ncc: Version 0.22.3
ncc: Compiling file index.js
(node:75823) DeprecationWarning: Chunk.modulesIterable: Use new ChunkGraph API
(node:75823) DeprecationWarning: Chunk.entryModule: Use new ChunkGraph API
 63kB  dist/licenses.txt
811kB  dist/index.js
```
That is being tracked [here](https://github.com/xz64/license-webpack-plugin/issues/67) and is hopefully resolved so we can minimize the output.

I did some local testing, but did not add any unit tests, let me know if there's some testing that would be worthwhile to add, and if you could provide some guidance on how to do so, I wasn't able to find a `contributing.md` or documentation.


src/index.js
```diff
@@ -206,68 +278,7 @@ module.exports = (
         }
       ]
     },
-    plugins: [
-      {
-        apply(compiler) {
```
thboop, line 211: this was just moved above so we can dynamically add the `LicenseWebpackPlugin` to this array based on the results of the `license` flag

src/cli.js
```diff
@@ -137,6 +138,8 @@ async function runCmd (argv, stdout, stderr) {
       "-q": "--quiet",
       "--watch": Boolean,
       "-w": "--watch",
+      "--license": Boolean,
+      "-l": "--license",
```
styfle, line 142: Thanks for the PR! 🎉 

Let's change `--license` to be a string so that the use can configure the the path to the output license file.

For example `ncc --license LICENSE.txt`.

Then you can pass this path to `outputFilename`

thboop: It is relative to the dist directory.

Sounds like a good change, I'll make this update today! Thanks for the review!

src/index.js
```diff
@@ -43,7 +44,8 @@ module.exports = (
     filterAssetBase = process.cwd(),
     quiet = false,
     debugLog = false,
-    transpileOnly = false
+    transpileOnly = false,
+    license = ""
```
styfle, line 48:
```suggestion
    license = ''
```

styfle:
> Currently, there is an open issue where that plugin outputs some warnings

Looks like the deprecation warnings were fixed in 2.3.3

https://github.com/xz64/license-webpack-plugin/issues/67#issuecomment-750391077

PR #579, opened 2020-08-20
Upgrade Webpack from 5.0.0-alpha.17 to 5.0.0-beta.28

guybedford:
This upgrades Webpack 5 from 5.0.0-alpha.17 to 5.0.0-alpha.23 which is enough to get the now-cli to build with ncc.

The upgrade path is quite difficult due to the use of experimental and internal APIs as well as the fact that there is a lot of movement on these alpha releases.

If CI passes here I'm going to attempt to progressively upgrade as far as I can as that would obviously be preferable.

//cc @styfle


scripts/build.js
```diff
@@ -68,7 +68,7 @@ async function main() {
 
   // detect unexpected asset emissions from core build
   function checkUnknownAssets (buildName, assets) {
-    assets = assets.filter(name => !name.endsWith('.cache') && !name.endsWith('.cache.js') && name !== 'LICENSES.txt');
+    assets = assets.filter(name => !name.endsWith('.cache') && !name.endsWith('.cache.js') && !name.endsWith('LICENSES.txt'));
```
styfle, line 71: Why change this?

guybedford:
Ok it looks like 5.0.0-beta.8 is about as far as we can go for now, since beta.9 enables tree shaking which seems to break some relocation stuff and I'm not sure how to turn it off.

scripts/build.js
guybedford: It ended up emitting a new LICENSES.txt asset, possibly from the webpack build itself.

guybedford:
I believe the tests should be passing now, ready for review.

scripts/build.js
```diff
@@ -17,7 +19,7 @@ async function main() {
       filename: "cli.js",
       externals: ["./index.js"],
       license: 'LICENSES.txt',
```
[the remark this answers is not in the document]
guybedford: I didn't follow the license PR merge and how that works. Happy to look into this separately to ensure it's meeting the requirements, but I don't think that falls under the scope of this PR.

guybedford:
@sokra it seems like we have quite a significant performance drop here on the Webpack 5 upgrade path, despite disabling a lot of the tree-shaking functionality as discussed. Are there any other options that might help here?

guybedford:
@styfle could you clarify if these numbers are before or after running `ncc cache clean` or doing multiple runs to fill the cache?

styfle:
@guybedford Those are three consecutive runs. I actually did many more but three got the point across that it was slower. I didn't do `ncc cache clean`, however.

guybedford:
@styfle a version of these benchmarks with a `ncc cache clean` before each run would indicate if the problem lies with the caching or with the actual operation.

scripts/build.js
```diff
@@ -32,29 +34,31 @@ async function main() {
       // chokidar, which is quite convenient
       externals: ["chokidar"],
```
sokra, line 35: chokidar is not a webpack dependencies since webpack 5. The watching is JS only and should bundle fine.

src/index.js
```diff
@@ -204,11 +176,20 @@ module.exports = (
       minimize: false,
       moduleIds: 'deterministic',
       chunkIds: 'deterministic',
-      mangleExports: false
+      mangleExports: false,
+      concatenateModules: false,
+      innerGraph: false,
+      sideEffects: false
```
sokra, line 182: 😢

src/index.js
```diff
@@ -366,11 +347,19 @@ module.exports = (
...
         if (source.startsWith('./'))
-          source = source.substr(2);
+          source = source.slice(2);
+        if (source.startsWith('(webpack)'))
+          source = 'webpack' + source.slice(9);
         if (source.startsWith('webpack/'))
-          return '/webpack/' + source.substr(8);
+          return '/webpack/' + source.slice(8);
         return sourceMapBasePrefix + source;
       });
```
sokra, line 364: This looks really weird. Have you looked into the `output.devtoolModule(Fallback)FilenameTemplate`, which allows to generate sources like you want to.

src/index.js
guybedford: Correctness is more important than bundle size.

src/index.js
guybedford: It's really not too much of a problem to have a custom map, but will bear that in mind for future reference, thanks.

guybedford:
@styfle there are a few things that are very important to ensure the correct function here, which you may not have done:

* `ncc cache clean` must be run between each version change
* `--external update-notifier` reduces the startup time significantly
* also, as per the build script, node_modules/fsevents should be removed if present

With the above, I tried four different runs here:

1. Currently used ncc@: 7695KB, startup 490ms
2. Latest ncc@0.23.0 (with bug): 7695KB, startup 490ms
3. This Webpack upgrade: 7840KB, startup 504ms
4. This Webpack upgrade with tree shaking options enabled: 7799KB, startup 505ms

The above does show a slight slowdown with the latest Webpack 5 upgrade here, but not one that seems too significant to me. I would assume this difference comes down to subtle interop / wrapper function changes in Webpack, but perhaps Tobias can share more here.

I've gone ahead and reenabled the treeshaking options.

The remaining practical question, unless Tobias has further insights, is just whether a 3% increase in startup time is a genuine concern and if we want to not upgrade either project based on that. Personally I don't see it as significant, but it's certainly not my decision.

scripts/build.js
guybedford: I've removed this.

src/index.js
guybedford: I actually went ahead and reenabled these options and all the tests seem to pass. I suppose I am familiar with Rollup treeshaking having correctness issues, apologies for the webpack projection :)

styfle:
@guybedford The startup time I mentioned earlier is not about `ncc` itself but about `vercel`, the code compiled by `ncc`.

1. `ncc cache clean` didn't help, I'm still seeing 1.2s - 1.3s startup time for `vercel`
2. `--external update-notifier` is already used as seen here: https://github.com/vercel/vercel/blob/bd706beba5c8022a41e78ad6b3c303f3a6726272/packages/now-cli/scripts/build.ts#L52-L57
3. `fsevents` is not included, please ignore this part since it was a typo on my end

To reproduce the slowdown, please try the following:

```sh
## Build NCC
cd /tmp
git clone https://github.com/vercel/ncc
cd ncc
git checkout webpack-upgrade
yarn
yarn build
/tmp/ncc/dist/ncc/cli.js cache clean

## Build Vercel CLI with the local NCC
cd /tmp
git clone https://github.com/vercel/vercel
cd vercel
sed -i "" "s|@zeit/ncc|/tmp/ncc/dist/ncc/cli.js|g" ./packages/now-cli/scripts/build.ts
yarn
yarn build

## Run Vercel CLI and time it
time ./packages/now-cli/dist/index.js --version
```

guybedford:
@styfle the numbers are above are exactly for the Vercel CLI build.

guybedford:
(that is those sizes and ms counts are for `./dist/index.js --version` in the CLI)

guybedford:
Which version of Node.js are you running?

styfle:
I'm using Node 12.18.3

guybedford:
@styfle I dug into this and it looks like the source map is significantly reducing the execution time in Node.js - if you build without the sourcemap, the execution time reduces to 500ms. I would advise turning off the sourcemap entirely for this reason, possibly turning off minification as the bandwidth saving isn't really worth it there either.

guybedford:
(that is - because the source map size increased, the execution time decreased!)

styfle:
This made ncc much smaller 👍 

- ncc@0.23.0 [install size](https://packagephobia.com/result?p=@vercel/ncc@0.23.0)
- ncc@0.24.0 [install size](https://packagephobia.com/result?p=@vercel/ncc@0.24.0)

3cp:
I got an error in compiled result with v0.24.0, but v0.23.0 works fine. I have not found any related issue in vercel/ncc, so I am not sure why it failed. But clearly v0.23.0 works.

https://github.com/ghiscoding/excel-builder.js/pull/1

The error:
https://github.com/ghiscoding/aurelia-slickgrid-demos/issues/2#issuecomment-691694363

guybedford:
@3cp if you can share the exact replication steps as a new issue that would help.

3cp:
@guybedford here is a demo for the bug. https://github.com/3cp/ncc-bug

I think this is technically a webpack bug. But it can be easily bypassed if ncc avoid the names of webpack global vars.

3cp:
New issue created.

PR #598, opened 2020-09-25
Adds support for PnP installs

arcanis:
This diff simply adds the PnpWebpackPlugin, which lets Webpack resolve dependencies when installed via PnP. Without it, all external dependencies fail to resolve and are silently kept as external requires.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "pdf2json": "^1.1.8",
     "pdfkit": "^0.8.3",
     "pg": "^7.6.1",
+    "pnp-webpack-plugin": "^1.6.4",
     "pug": "^2.0.3",
     "react": "^16.6.3",
     "react-dom": "^16.6.3",
@@ -108,8 +109,5 @@
     "vue-server-renderer": "^2.5.17",
     "webpack": "5.0.0-beta.28",
     "when": "^3.7.8"
-  },
-  "dependencies": {
-    "pnp-webpack-plugin": "^1.6.4"
   }
 }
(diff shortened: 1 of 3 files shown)


styfle (commented):
Thanks!

A couple things:

1. We need a test for this new behavior.
2. Seems like this should be opt-in with a flag? We recently had a [PR to `@vercel/nft`](https://github.com/vercel/nft/pull/153) and decided to add a hook to allow users to override the resolution behavior rather than use PnP as default behavior. 
3. This package says its built on `ts-pnp`. Does that only support TS or does it also support JS?

arcanis:
- Will check to add a test - it's a bit tricky to test PnP on non-PnP repos, but I'll see what I can do.

- The plugin has strictly no effect unless the project uses PnP, so in a sense it's already opt-in: it requires the user to explicitly use Yarn 2. Since ncc just doesn't behave well there without this patch, I think it should be done without CLI flag.

- TS-PnP is a dependency just for easier integration with fork-ts-webpack-checker (the plugin uses it to provide a few helpers). It's not used in this PR.

styfle:
Sounds good! If we could enable PnP for a subdirectory, that would be perfect for a new test 👌

arcanis:
Actually, funny story: ncc already supports PnP! I didn't notice that you were using a compatible Webpack 5 release, because I was working on the latest `@zeit/ncc` release ... instead of `@vercel/ncc` 😅 switching to the new name, now everything seems to work!

Thanks for linking the thread! 🍀

PR #605, opened 2020-10-17
fix: respect quiet flag on ts build

### LINKED ISSUE #603: Typescript build doesn't respect the `quiet` flag
When using the `build` command on Typescript files, there is some console output even when the `quiet` flag is passed.

```bash
jamiedavenport@Jamies-MBP  ~/ncc-test  ncc version
0.24.1
 jamiedavenport@Jamies-MBP  ~/ncc-test  ncc build --quiet index.ts
ncc: Using typescript@3.2.2 (ncc built-in)
 jamiedavenport@Jamies-MBP  ~/ncc-test  ncc build index.ts --quiet
ncc: Using typescript@3.2.2 (ncc built-in)
 jamiedavenport@Jamies-MBP  ~/ncc-test 
```

`index.ts`
```ts
console.log("Hello, World");
```

I believe the issue is [here](https://github.com/vercel/ncc/blob/master/src/typescript.js#L6-L13) where `console.log` should be wrapped in some kind of `if(!quiet)` check.

huozhi:
Fix #603

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -49,6 +49,9 @@ module.exports = (
     license = ''
   } = {}
 ) => {
+  process.env.__NCC_OPTS = JSON.stringify({
+    quiet
+  });
   const ext = extname(filename);
 
   if (!quiet) {
--- a/src/typescript.js
+++ b/src/typescript.js
@@ -1,14 +1,15 @@
 
 const { Module } = require('module');
 const m = new Module('', null);
+const {quiet} = JSON.parse(process.env.__NCC_OPTS || '{}');
 m.paths = Module._nodeModulePaths(process.env.TYPESCRIPT_LOOKUP_PATH || (process.cwd() + '/'));
 let typescript;
 try {
   typescript = m.require('typescript');
-  console.log("ncc: Using typescript@" + typescript.version + " (local user-provided)");
+  if (!quiet) console.log("ncc: Using typescript@" + typescript.version + " (local user-provided)");
 }
 catch (e) {
   typescript = require('typescript');
-  console.log("ncc: Using typescript@" + typescript.version + " (ncc built-in)");
+  if (!quiet) console.log("ncc: Using typescript@" + typescript.version + " (ncc built-in)");
 }
 module.exports = typescript;
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -90,21 +90,7 @@ async function main() {
 
   writeFileSync(__dirname + "/../dist/ncc/cli.js", cli, { mode: 0o777 });
   writeFileSync(__dirname + "/../dist/ncc/index.js", index);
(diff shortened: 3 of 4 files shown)


src/typescript.js
```diff
@@ -1,14 +1,17 @@
 
 const { Module } = require('module');
+
 const m = new Module('', null);
+const log = (...args) => { if (!quiet) { console.log(...args); } }
```
styfle, line 5: This function isn't used?

src/typescript.js
```diff
@@ -1,14 +1,17 @@
...
 m.paths = Module._nodeModulePaths(process.env.TYPESCRIPT_LOOKUP_PATH || (process.cwd() + '/'));
 let typescript;
 try {
   typescript = m.require('typescript');
-  console.log("ncc: Using typescript@" + typescript.version + " (local user-provided)");
+  if (!global.nccQuiet) console.log("ncc: Using typescript@" + typescript.version + " (local user-provided)");
```
styfle, line 10: Can this be solved without a global variable?

huozhi: another solution I can think of is: 
create a config file for CLI arguments/flags, can update the config when args parsed the arguments from command line. other files will use this config file to consume flags.

```
// index.js
import cliargs from './cliargs'

params = args.parse()
cliargs.update(params)

// typescript.js
import cliargs from './cliargs'

quite = cliargs.get('quiet')
```

if you feel this is proper I'll update the PR later

src/typescript.js
huozhi: good catch, i'll remove it in next commit.

src/index.js
```diff
@@ -49,8 +50,11 @@ module.exports = (
     license = ''
   } = {}
 ) => {
+  cliArgs.consume({
+    'quiet': quiet
+  });
   const ext = extname(filename);
-
+  global.nccQuiet = quiet;
```
styfle, line 57: The global is unused now

styfle (commented):
We also need a test so we dont regress

styfle (dismissed):
Great, thanks!

huozhi:
one extra question not related to this PR but I encounter when I test locally: how could I build the latest ncc changes to dist? it seems `yarn build` doesn't generate the latest code from src to dist. did I miss any steps?

huozhi:
@styfle this solution seems not working. the cliargs that typescript.js file get is always empty, but the `cliargs.consume` did get called before it. but I don't know why..

will adding a new env var like `process.env.NCC_QUIET` be acceptable?

styfle:
@huozhi Let's use `process.env.__NCC_OPTS = JSON.stringify(opts)` and then the appropriate files can deserialize and use any necessary options

styfle (approved):
Great, thanks!

PR #614, opened 2020-11-06
Allow passing the webpack target

ijjk:
This is a follow-up to https://github.com/vercel/next.js/pull/18873 to allow us to pass a target of `node,es5` to `ncc` since `webpack` defaults to using `es6` code for `node` which breaks ie11 compatibility

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -35,7 +35,7 @@ Options:
   --v8-cache               Emit a build using the v8 compile cache
   --license [file]         Adds a file containing licensing information to the output
   --stats-out [file]       Emit webpack stats as json to the specified output file
-  --webpack-target         What build target to use for webpack (default: node)
+  --target                  What build target to use for webpack (default: es6)
 `;
 
 // support an API mode for CLI testing
@@ -148,7 +148,7 @@ async function runCmd (argv, stdout, stderr) {
       "-t": "--transpile-only",
       "--license": String,
       "--stats-out": String,
-      "--webpack-target": String
+      "--target": String
     }, {
       permissive: false,
       argv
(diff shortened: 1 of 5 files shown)


styfle (approved):
Looks good, thanks!

src/cli.js
```diff
@@ -35,6 +35,7 @@ Options:
   --v8-cache               Emit a build using the v8 compile cache
   --license [file]         Adds a file containing licensing information to the output
   --stats-out [file]       Emit webpack stats as json to the specified output file
+  --webpack-target         What build target to use for webpack (default: node)
```
styfle, line 38: Lets call this `--target` and accept `es5` so we can always append `node`

styfle: We can validate the input so that it must begin with `es` like `es5` or `es2020`

ijjk: Updated to accept es values and to always append `node`

src/cli.js
```diff
@@ -35,7 +35,7 @@ Options:
   --v8-cache               Emit a build using the v8 compile cache
   --license [file]         Adds a file containing licensing information to the output
   --stats-out [file]       Emit webpack stats as json to the specified output file
-  --webpack-target         What build target to use for webpack (default: node)
+  --target                  What build target to use for webpack (default: es6)
```
styfle, line 38: Where do you see default is `es6`? Is that documented by webpack somewhere?

styfle (approved):

src/cli.js
ijjk: It's the default for webpack `v5` with `node` https://github.com/webpack/webpack/blob/7415a618460795874fdf95c8f03363ea96709d52/lib/config/target.js#L163-L171

PR #627, opened 2020-12-15
fix: correct declaration output dir

### LINKED ISSUE #481: Wrong output destination for declarations
It generates

```
yarn run v1.17.3
$ /home/charlike/github/tunnckoCore/opensource/node_modules/.bin/ncc build src/index.ts
ncc: Version 0.20.5
ncc: Compiling file index.js
ncc: Using typescript@3.6.4 (local user-provided)
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/test/index.d.ts
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/test/index.d.ts.map
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/utils.d.ts.map
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/index.d.ts
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/index.d.ts.map
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/plugins/props.d.ts.map
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/plugins/params.d.ts
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/plugins/params.d.ts.map
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/plugins/initial.d.ts
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/plugins/initial.d.ts.map
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/plugins/body.d.ts
  0kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/plugins/body.d.ts.map
  1kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/utils.d.ts
  1kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/types.d.ts
  1kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/types.d.ts.map
  1kB  dist/home/charlike/github/tunnckoCore/opensource/packages/parse-function/dist/typings/src/plugins/props.d.ts
346kB  dist/index.js
350kB  [5235ms] - ncc 0.20.5
Done in 5.52s.
```

The tsconfig inside the package root is

```json
{
  "$schema": "http://json.schemastore.org/tsconfig",
  "compilerOptions": {
    "declarationDir": "./dist/typings",
    "typeRoots": [
      "./node_modules/@types",
      "../../node_modules/@types",
      "../../@types"
    ],
    "skipLibCheck": true,
    "alwaysStrict": true,
    "newLine": "LF",
    "declaration": true,
    "declarationMap": true,
    "esModuleInterop": true,
    "experimentalDecorators": true,
    "emitDecoratorMetadata": false,
    "allowSyntheticDefaultImports": true,
    "forceConsistentCasingInFileNames": true,
    "importHelpers": true,
    "incremental": false,
    "jsx": "react",
    "lib": ["dom", "esnext"],
    "module": "esnext",
    "moduleResolution": "node",
    // "noEmit": true,
    "strict": true,
    "sourceMap": true,
    "stripInternal": true,
    "noUnusedLocals": true,
    "noImplicitReturns": true,
    "resolveJsonModule": true,
    "noUnusedParameters": true,
    "target": "esnext",
    "rootDir": "src",
    "outDir": "dist",
    "baseUrl": "."
  },
  "exclude": ["test"]
}
```

With the `tsc` it is correct `./dist/typings`

The current workaround is

```json
{
  "scripts": {
    "prebuild": "rm -rf dist",
    "build": "ncc build src/index.ts -sm -o dist/main",
    "postbuild": "mv \"dist/main/$PWD/dist/typings\" dist && rm -rf dist/main/home"
  }
}
```

zeroooooooo:
If declarationDir is special, the output dir will be unexpected, fixes #481, and I run typescript example will cause same error as follow
image

`tsconfig.json` as follow
image

I give a quick fix, rewrite dir before write file into disk, seems can fix the problem.


guybedford:
Just to clarify the expected behaviour here - do you need these types to be included in the output?

If so, the fix looks like the right sort of approach to me, but I would make two suggestions to get this merged:

1. Include a `path.isAbsolute` check along with the `.endsWith('.d.ts')` check, and treat this purely as a renormalization based on the base directory, rather than an ad hoc folder split (which will fail on arbitrary subfolder structures).
2. Move this logic into `src/index.js` where the `finalizeHandler` does asset completions.

styfle:
Additionally, we'll need a test for this new behavior

zeroooooooo:
> Just to clarify the expected behaviour here - do you need these types to be included in the output?

I think this may not be a common use case, I use this is because I need to compile code into multiple types of modules(es, commonjs), and they shared the tsconfig. to avoid duplicate generate declarations, it needs to compile declarations into a separate folder. so I make this changes, but it seems not necessary if use `ncc` alone.

> If so, the fix looks like the right sort of approach to me, but I would make two suggestions to get this merged:
> 
> 1. Include a `path.isAbsolute` check along with the `.endsWith('.d.ts')` check, and treat this purely as a renormalization based on the base directory, rather than an ad hoc folder split (which will fail on arbitrary subfolder structures).
> 2. Move this logic into `src/index.js` where the `finalizeHandler` does asset completions.

I move logic into `src/index.js`, but make some changes, you can check if it is ok.

zeroooooooo:
> Additionally, we'll need a test for this new behavior

test added

guybedford:
Thanks for the test refinements here.

I'm still not super comfortable with assuming that the type folder and `outDir` mean the same thing. Ideally we should really read the tsconfig.json value out and use that as the reference point rather than requiring this separate option.

zeroooooooo:
> Thanks for the test refinements here.
> 
> I'm still not super comfortable with assuming that the type folder and `outDir` mean the same thing. Ideally we should really read the tsconfig.json value out and use that as the reference point rather than requiring this separate option.

Yes, it should read 'outDir' from tsconfig. It adds the outDir path as a prefix to the asset in the final process logic in 'cli'. Js', I thought it would affect the result, so I used 'outDir' directly, but find out it won't. Now I use outDir from tsconfig to handle the logic.

Thanks for the detailed review. You can check it again to see if the appropriate changes have been made.

guybedford (approved):
This looks great to me now, thanks for working on a comprehensive approach here.

styfle (approved):
Thanks!

PR #631, opened 2020-12-23
Enable minification for sourcemap-register.js

xom9ikk:
File `scripts/build.js` contains invalid configuration for `ncc`.
Most likely it does not do what was expected. After fixed, the `/dist/ncc/sourcemap-register.js.cache.js` file becomes minified. And its weight is reduced by `86kb`.

**File:** `/dist/ncc/sourcemap-register.js.cache.js`
**Original size:** `127kb`
**Fixed:** `41kb`

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -62,7 +62,7 @@ async function main() {
 
   const { code: sourcemapSupport, assets: sourcemapAssets } = await ncc(
     require.resolve("source-map-support/register"),
-    { filename: "sourcemap-register.js", minify: true, v8cache: true }
+    { filename: "sourcemap-register.js", minify, v8cache: true }
   );
   checkUnknownAssets('source-map-support/register', Object.keys(sourcemapAssets));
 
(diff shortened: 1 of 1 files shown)


scripts/build.js
```diff
@@ -62,7 +62,7 @@ async function main() {
 
   const { code: sourcemapSupport, assets: sourcemapAssets } = await ncc(
     require.resolve("source-map-support/register"),
-    { filename: "sourcemap-register.js", minfiy: true, v8cache: true }
+    { filename: "sourcemap-register.js", minify: true, v8cache: true }
```
styfle, line 65:
```suggestion
    { filename: "sourcemap-register.js", minify, v8cache: true }
```

styfle (commented):
Nice catch! Let's use the existing `minify` variable like the others above

styfle (approved):

xom9ikk:
@styfle yes you are right!)

PR #648, opened 2021-02-01
chore: add console output when minifying fails

Songkeys:
It took me quite a while to find that my output is not minified because there's a "formidable" dependency that causes this stage failing, not because I passed a wrong flag. I think there should be a console output to let users know the output is not minified because of failure.


src/index.js
```diff
@@ -410,11 +410,14 @@ function ncc (
...
       // For some reason, auth0 returns "undefined"!
       // custom terser phase used over Webpack integration for this reason
-      if (result.code !== undefined)
+      if (result.code !== undefined) {
         ({ code, map } = {
           code: result.code,
           map: sourceMap ? JSON.parse(result.map) : undefined
         });
+      } else {
+        console.log('Error happens when minifying. The result will not be minified.')
```
styfle, line 419:
```suggestion
        console.log('An error occurred while minifying. The result will not be minified.')
```

styfle (approved):
Thanks!

styfle:
We probably just need to update `terser` to handle the case when its failing to minify.

Would you want to submit another PR to do that?

Songkeys:
@styfle Sorry, no. I've moved to vercel/nft 😅. And I think a newer version cannot solve this issue either, right?

PR #654, opened 2021-02-22
Support Regexp externals

### LINKED ISSUE #647: Support for wildcard externals
`ncc` should allow a user to define an external that's a wildcard, such that they do not need to manually inspect a bundled dependency for all of its sub-imports.

For example, bundling `browserslist` requires all of these externals:
```
caniuse-lite
caniuse-lite/dist/unpacker/feature
caniuse-lite/dist/unpacker/region
caniuse-lite/dist/unpacker/agents
```

It'd be much better to just specify `caniuse-lite`!

huozhi:
Resolves #647

### Changes

Support regex pattern for externals, to exclude more external libs easily.
Regexp is supported natively, this can avoid adding extra libraries to support wildcard patterns like `pattern*`.

### Usage

```js
 {
   externals: {
     'externaltest': 'externalmapped',
     '/\\w+-regex/': 'regexexternal',
   }
}
```


test/index.test.js
```diff
@@ -42,7 +42,8 @@ for (const unitTest of fs.readdirSync(`${__dirname}/unit`)) {
       },
       externals: {
         'piscina': 'piscina',
-        'externaltest': 'externalmapped'
+        'externaltest': 'externalmapped',
+        'regexexternal': /\w+-regex/,
```
guybedford, line 46: I think it would make more sense to reverse this around so that the key is always the match and the value is always the replacement. Using the value as the replacement in the regex case seems odd otherwise.

So:

```js
  '/\w+-regex/': 'regexexternal-$1': 
```

It might even be nice to permit match replacements on the RHS as well.

guybedford (commented):
This looks great to me, sorry for the delayed review on this @huozhi.

The implementation seems sensbile. See the comment re the matching form.

test/index.test.js
guybedford: I went ahead and added this switch, but didn't add replacements for now.

guybedford (approved):
Rebased, and changed the ordering in the object form.

Thanks again @huozhi for your work here.

huozhi:
Thank you for the great advice and quick improvements! @guybedford

styfle (approved):
Thanks!

PR #665, opened 2021-03-18
"exports" conditions semantics

guybedford:
This updates ncc to follow the "node" condition and not the "module" condition in the exports field with tests.

This will ensure that the built semantics exactly mirror the Node.js semantics.

Support for the "imports" field similarly is pending https://github.com/webpack/webpack/issues/12928.


styfle (approved):

src/index.js
```diff
@@ -52,8 +52,24 @@ function ncc (
     transpileOnly = false,
     license = '',
     target,
+    production = true,
```
styfle, line 55: Does node automatically consider 'production' exports?

styfle: Seems like this should be `conditions: string[]` instead

PR #667, opened 2021-03-19
Interop test case

guybedford:
This is a failing test case for one of the few of the remaining module interop cases that is different between Webpack and Node.js.

I'm hopeful we can get an upstream fix for this in Webpack in some form.

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -1,5 +1,9 @@
 [
   {
+    args: ["run", "test/fixtures/interop-test.js"],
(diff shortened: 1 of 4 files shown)


styfle (approved):

guybedford:
It turns out I spoke too soon here and @sokra has done some amazing work smoothing these edge cases (also per https://sokra.github.io/interop-test/).

As long as we are in the strict interop mode for Webpack there is full Node.js compat on the API patterns now.

PR #671, opened 2021-03-21
TS interop test and fix

guybedford:
This fixes the TypeScript interop including a fix as well for enabling the synthetic default imports option and outputting ESM for Webpack to apply its own interop patterns instead of relying on the more complex TypeScript interop output. This will also reduce the size of TypeScript applications with ncc.


src/index.js
```diff
@@ -299,6 +299,8 @@ function ncc (
               transpileOnly,
               compiler: eval('__dirname + "/typescript.js"'),
               compilerOptions: {
+                allowSyntheticDefaultImports: true,
```
styfle, line 302: Do we also want to use `esModuleInterop: true`?

styfle (approved):

src/index.js
guybedford: I'm pretty sure that only affects output (to CJS, which this is now not), not interpretation.

PR #252, opened 2019-01-29
Local TS integration test

guybedford:
This follows #251 and adds a full test for the local TypeScript version lookup.

PR #674, opened 2021-03-23
enhance: upgrade local ts

### LINKED ISSUE #606: Fails to compile when tsconfig has `incremental: true`
Output:

```
ncc build src/index.ts
ncc: Version 0.24.1
ncc: Compiling file index.js
ncc: Using typescript@4.0.3 (local user-provided)
Error: [tsl] ERROR
      TS5074: Option '--incremental' can only be specified using tsconfig, emitting to single file or when option `--tsBuildInfoFile` is specified.
    at /Users/gozala/.config/yarn/global/node_modules/@vercel/ncc/dist/ncc/index.js.cache.js:1:3198784
    at /Users/gozala/.config/yarn/global/node_modules/@vercel/ncc/dist/ncc/index.js.cache.js:1:2305642
    at eval (eval at create (/Users/gozala/.config/yarn/global/node_modules/@vercel/ncc/dist/ncc/index.js.cache.js:1:1291515), <anonymous>:11:1)
```

Reproducible on the following tree https://github.com/Gozala/typescript-error-reporter-action/tree/a103de6e96e3797e40b455e878b25e8cf3a8b2d1

huozhi:
Fixes #606

* bump ts-loader version
* bump built-in ts version
* override `incremental` option to `false`


styfle (commented):
Thanks! Looks like CI is failing though.

huozhi:
@styfle @guybedford, if I want to upgrade the ts lib assets generated in `test/integration/node_modules/typescript/`, (introduced #252) what's the proper coommand I should run? I tried `./dist/ncc/cli.js build src/typescript.js -o test/integration/node_modules/typescript/index.js -m` but it emits lots of assets

```
index.js typescript  typescript1 typescript2 typescript3 typescript4
```

guybedford:
@huozhi that sounds right to me, but TypeScript is a very complex build for ncc already - just updating TypeScript is a job unto itself. Is it working?

yarn.lock
```diff
@@ -8988,13 +8978,13 @@ loader-runner@^4.2.0:
   integrity sha512-92+huvxMvYlMzMt0iIOukcwYBFpkYJdpl2xsZ7LrlayO7E8SOv+JJUEK17B/dJIHAOLMfh2dZZ/Y18WgmGtYNw==
 
 loader-utils@^1.0.2:
-  version "1.1.0"
-  resolved "https://registry.yarnpkg.com/loader-utils/-/loader-utils-1.1.0.tgz#c98aef488bcceda2ffb5e2de646d6a754429f5cd"
-  integrity sha1-yYrvSIvM7aL/teLeZG1qdUQp9c0=
+  version "1.4.0"
+  resolved "https://registry.npmjs.org/loader-utils/-/loader-utils-1.4.0.tgz#c579b5e34cb34b1a74edc6c1fb36bfa371d5a613"
```
styfle, line 8982: Something seems wrong here. Why was `registry.yarnpkg.com` changed to `registry.npmjs.org`?

styfle (commented):
I'm not sure if a symlink for `test/integration/node_modules/typescript/index.js ` is correct here. What do you think @guybedford ?

guybedford:
@styfle the symlink approach seems fine to me actually. Note we should be sure to thoroughly test the TypeScript upgrade here manually before releasing, and it is a major release to upgrade TypeScript.

styfle:
@guybedford All the CI tests are passing. Any other TS tests you have in mind?

guybedford:
@styfle we may be missing test coverage of the **bundled** typescript version. That is, we should add a test that does a similar symlink trick, but to `./dist/typescript` instead. Because otherwise `node_modules/typescript` will always be used as a local TypeScript always takes precedence so does in these tests as well.

yarn.lock
huozhi: my local config lead to this unexpected change. will roll it back to use yarnpkg registry

huozhi (commented):
this ts upgration is truly a big gap (from 3.2.2 to 3.9.9). 
one idea is to only upgrade to 3.4 to support `incremental` option, the gap is small.
another option is to change purpose of this PR to "upgrade local provided ts" and release later.

which one do you prefer?

styfle (approved):

huozhi:
I'm kinda unclear about the approach of testing `./dist/typescript.js` with symlink, would you mind clarifying it a bit more specific? I can work on that then. Thanks

guybedford:
@huozhi perhaps a symlink is the wrong direction actually - instead we just want to **not** load the local `test/node_modules/typescript` by ensuring that TypeScript cannot be located.

One way to do this would be to construct a tmp folder at the system root - `/tmp/s9d8f0s9d8/test.ts` and then execute that ts file. Because there is no `node_modules` containing `typescript` from such a path, this will force ncc to use its local inbuilt typescript version and thus test that.

Alternatively, simply running ncc from a difference cwd could also do this:

```
cd /
ncc run /path/to/ncc/test/unit/ts-ext/test.ts
```

Tests like the above might be possible to construct with the `test/cli.js` file (https://github.com/vercel/ncc/blob/master/test/cli.js) by adding a `cwd` option support to that test system.

So that's two possible ways to test this, basically if ncc finds a `node_modules/typescript` in cwd, it uses it, and notes that it is in the console. And because `ncc/node_modules` and `ncc/test/node_modules` both have local typescript versions, we are never testing the case where it isn't found. If you can think of another way apart from my above two suggestions to catch this other case feel free to try something else too and just let me know if I can clarify further at all.

huozhi:
I use the `TYPESCRIPT_LOOKUP_PATH` env as override option for tests. If the folder is invalid ncc can fallback to built-in ts lib. hope it's sth you're expecting.

guybedford (approved):
Looks great, thanks for your work on this!

test/index.test.js
```diff
@@ -84,7 +84,8 @@ for (const unitTest of fs.readdirSync(`${__dirname}/unit`)) {
 for (const cliTest of eval(fs.readFileSync(__dirname + "/cli.js").toString())) {
   it(`should execute "ncc ${(cliTest.args || []).join(" ")}"`, async () => {
     const ps = fork(__dirname + (coverage ? "/../src/cli.js" : "/../dist/ncc/cli.js"), cliTest.args || [], {
-      stdio: "pipe"
+      stdio: "pipe",
+      env: cliTest.env,
```
styfle, line 88: Should this extend `process.env`?

Something like `{...process.env, ...cliTest.env}`

styfle (approved):
Great work, thanks! 🎉

PR #685, opened 2021-04-07
fix:  unknown compiler option 'incremental'

Songkeys:
image

When using a lower version of typescript, ts-loader will argue that there is an unknown compiler option `incremental`.


Songkeys:
cc @huozhi from #674

huozhi (approved):

styfle (approved):

Issue #689, opened 2021-04-10
SyntaxError: Unexpected token '}' on windows


UziTech:
Upgrading from v0.28.0 to v0.28.2 in a github action

From the [log](https://github.com/UziTech/action-setup-atom/pull/186/checks?check_run_id=2310597856):

```
D:\a\action-setup-atom\action-setup-atom\dist\index.js:5239
/******/ })()
         ^

SyntaxError: Unexpected token '}'
    at Module._compile (internal/modules/cjs/loader.js:895:18)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:995:10)
    at Module.load (internal/modules/cjs/loader.js:815:32)
    at Function.Module._load (internal/modules/cjs/loader.js:727:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1047:10)
    at internal/main/run_main_module.js:17:11
```

This seems to only happen on Microsoft Windows Server 2019 (currently windows-latest on github actions). I tested it on Window 10 and it ran just fine.

styfle:
> This seems to only happen on Microsoft Windows Server 2019 (currently windows-latest on github actions). I tested it on Window 10 and it ran just fine.

That is odd. Can you confirm the Node.js version is the same on both? Perhaps its an outdated version in GH Actions?

UziTech:
Node 14 on github-actions (where it fails). Node 15 on windows 10 (where it passes)

styfle:
I created PR #691 to see if I could reproduce with CI but it doesn't get past `node-gyp`.

If you want to fix windows, PRs are welcome 🙂

UziTech:
> it doesn't get past `node-gyp`.

Maybe this is the issue I am running into.

I can confirm that it works in v0.28.0 but not in v0.28.1+

UziTech:
After much debugging I found that running the file output by @vercel/ncc (even in github actions windows-latest) it runs just fine, but when running the output as an action in github actions it fails with the above error (only on windows-latest).

It seems like this might be an issue with something that github actions injects into the output file.

UziTech:
I tried building with @vercel/ncc manually to see if I could confirm it was github actions but the issue seems to only present itself when the build and run happen on github actions. I am at a loss for what to try next. I feel like github actions is doing something behind the scenes but I cannot figure out what.

Build and run on github actions fails:
[...]

PR #691, opened 2021-04-11
chore: windows ci

styfle:
Related to #689

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "mkdirp": "^0.5.1",
     "mongoose": "^5.3.12",
     "mysql": "^2.16.0",
-    "node-gyp": "^3.8.0",
+    "node-gyp": "^8.0.0",
     "npm": "^6.13.4",
     "oracledb": "^4.2.0",
     "passport": "^0.4.0",
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
     "build": "node scripts/build",
     "build-test-binary": "cd test/binary && node-gyp rebuild && cp build/Release/hello.node ../integration/hello.node",
     "codecov": "codecov",
-    "test": "node --expose-gc --max_old_space_size=3072 node_modules/.bin/jest",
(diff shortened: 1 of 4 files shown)


guybedford:
Perhaps try updating the version of `node-gyp` here?

styfle:
@guybedford Bumping node-gyp helped but now jest doesn't run. Seems like we need to expose GC and make it windows compatible in order to run the tests.

package.json
```diff
@@ -15,8 +15,8 @@
     "build": "node scripts/build",
     "build-test-binary": "cd test/binary && node-gyp rebuild && cp build/Release/hello.node ../integration/hello.node",
     "codecov": "codecov",
-    "test": "node --expose-gc --max_old_space_size=3072 node_modules/.bin/jest",
-    "test-coverage": "node --expose-gc --max_old_space_size=3072 node_modules/.bin/jest --coverage --globals \"{\\\"coverage\\\":true}\" && codecov",
```
guybedford, line 19: `--max_old_space_size` is necessary here.

styfle: Thanks! I added cross-spawn and that got past the memory issue, but now jest doesn't find any tests on Windows 🤔

guybedford:
Seems like it might have something to do with the first `?` in the configuration matcher - https://github.com/vercel/ncc/blob/master/jest.config.js#L5.

CI reports: `testMatch: D:/a/ncc/ncc/test\?(*.)+(spec|test).js?(x) - 0 matches`

While the matcher configuration has: `testMatch: ["<rootDir>/test/?(*.)+(spec|test).js?(x)"]`

Try simplifying the rule to just:

```
testMatch: ["<rootDir>/test/*.test.js"]
```

styfle:
Good call!

Looks like some of the tests are failing because windows emits `\r\n` instead of the expected `\n` so thats not a big issue.

However, there are a couple that look like actual bugs: https://github.com/vercel/ncc/pull/691/checks?check_run_id=2333639573#step:9:318

styfle:
Fixed in #896

Issue #690, opened 2021-04-10
ncc builds with old version of dependency


tonyhallett:
I built with a dependency that had, in error, imported a module relative to src folder. I noticed the error when the javascript failed.
I corrected the dependency and updated the version in my project.
I built again - same javascript

ncc cache clean - correct javascript

How is the cache supposed to work ?

timfish:
I've found something similar with ` @vercel/ncc@0.28.3`.

Fixed an issue and spent a while trying to work out why I was still getting the same error.

PR #698, opened 2021-04-27
Adjust caching options

guybedford:
This updates the Webpack caching options to disable `node_modules` managed paths caching by default and to always use hashes instead of timestamps.

This should hopefully ensure better cache integrity for ncc builds and resolve https://github.com/vercel/ncc/issues/690.


styfle (approved):
Thanks!

Issue #697, opened 2021-04-22
Invalid assignment on 0.28.x (superagent as dependency)


talaikis:
Works fine with <= 0.27.0, when 0.28.x does not build one file, that has superagent as dependency:

```bash
...\node_modules\@vercel\ncc\dist\ncc/index.js.cache.js:20
<code>
JS_Parse_Error [SyntaxError]: Invalid assignment
    at js_error (\node_modules\@vercel\ncc\dist\ncc/index.js.cache.js:20:163522)
    at croak (\node_modules\@vercel\ncc\dist\ncc/index.js.cache.js:20:174584)
    at maybe_assign (\node_modules\@vercel\ncc\dist\ncc/index.js.cache.js:20:204547)
    at expression (\node_modules\@vercel\ncc\dist\ncc/index.js.cache.js:20:204656)
    at simple_statement (\node_modules\@vercel\ncc\dist\ncc/index.js.cache.js:20:179275)
    at statement (\node_modules\@vercel\ncc\dist\ncc/index.js.cache.js:20:176763)
    at _embed_tokens_wrapper (\node_modules\@vercel\ncc\dist\ncc/index.js.cache.js:20:175715)
    at if_ (\node_modules\@vercel\ncc\dist\ncc/index.js.cache.js:20:187029)
    at statement (\node_modules\@vercel\ncc\dist\ncc/index.js.cache.js:20:177758)
    at _embed_tokens_wrapper (\node_modules\@vercel\ncc\dist\ncc/index.js.cache.js:20:175715) {
  filename: '0',
  line: 2338,
  col: 45,
  pos: 54545
}
```

guybedford:
Thanks for posting the issue - can you share some more details on the replication here? I tried the following build:

test.js
```js
console.log(require('superagent'));
```

with superagent@6.1.0, and `ncc run test.js` seems to work fine for me.

talaikis:
I see, ok, here;s full file then:

```js
// used @mailchimp/mailchimp_marketing
const superagent = require('superagent')
const querystring = require('querystring')
const { createHash } = require('crypto')

const md5 = (s) => {
  const md5 = createHash('md5')
  md5.update(s)
  return md5.digest('hex')
}

const config = {
  apiKey: '<removed>,
  server: 'us2'
}

module.exports.addSubsriber = async (email, fName, lName) => {
  await addListMember('<rm>', {
    email_address: email,
    status: 'subscribed',
    update_existing: true,
    merge_fields: {
[...]

Issue #693, opened 2021-04-14
Breaking change in version 0.28.0


EndBug:
Hi, I've found out that when I use version ^0.28.0 my GitHub action, which uses ncc to compile, is no longer able to require a file provided by GitHub.

Here is the code (TypeScript) that I'm using:
```ts
const eventPath = process.env.GITHUB_EVENT_PATH ?? '/github/workflow/event.json',
    event = eventPath && require(eventPath)
```

`GITHUB_EVENT_PATH` is an env variable provided by GitHub, and in my case its value is `/home/runner/work/_temp/_github_workflow/event.json`
`require(eventPath)` throws this error:
> Error: Cannot find module '/home/runner/work/_temp/_github_workflow/event.json'

I've verified that the file still exists and is still valid JSON, and that when I compile my current code with version 0.27.0 it runs with no issues.

guybedford:
@EndBug can you perhaps look at the generated code to see the difference between what is being output here? It's difficult to look into this without a full replication.

EndBug:
@guybedford Sure!
These are the two build files (I've turned off compression for now, since that would make it unreadable)
- 0.27.0: https://github.com/EndBug/add-and-commit/blob/682e840606b6548b292626282c96ddc1049d4d5f/lib/index.js#L11115
- 0.28.0: https://github.com/EndBug/add-and-commit/blob/6d00245624492ed4e9efa4c0df829668e1ae00ef/lib/index.js#L10912

Here is the part of compiled code that corresponds to the interested region:

```js
// 0.27.0, line 11115
const eventPath = process.env.GITHUB_EVENT_PATH, event = eventPath && require(eventPath) // ...

// 0.28.0, line 10912
const eventPath = process.env.GITHUB_EVENT_PATH, event = eventPath && __nccwpck_require__(875)(eventPath) // ...
```

Looking at these it seems like the `require` has been turned to some kind of webpack require, is that correct?

Songkeys:
Seems that it gets transpiled into a webpackEmptyContext: https://github.com/EndBug/add-and-commit/blob/6d00245624492ed4e9efa4c0df829668e1ae00ef/lib/index.js#L11066-L11077

Originally, `require(var)` won't be recognised by ncc since it might be a "unable-to-be-resolved" dynamic require, but the new webpack introduced in 0.28.0 looks like transforming it directly at compile time rather than leaving it there...?

I think you could change it into a more complex form to not make it recognisable as a temporary solution for now, like:

```javascript
[...]

PR #701, opened 2021-04-28
fix: publish should build without cache

styfle:
- Related to #697 
- Related to #693

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1,14 +1,19 @@
 const ncc = require("../src/index.js");
-const { statSync, writeFileSync, readFileSync, unlinkSync } = require("fs");
+const { statSync, writeFileSync, readFileSync, unlinkSync, rmdirSync } = require("fs");
 const { promisify } = require("util");
-const { relative } = require("path");
+const { relative, join } = require("path");
 const copy = promisify(require("copy"));
 const glob = promisify(require("glob"));
 const bytes = require("bytes");
 
 const minify = true;
+const v8cache = true;
+const cache = join(__dirname, '..', '.cache');
 
 async function main() {
+  if (process.argv[2] === '--no-cache') {
+    rmdirSync(cache, { recursive: true });
+  }
   for (const file of await glob(__dirname + "/../dist/**/*.@(js|cache|ts)")) {
     unlinkSync(file);
   }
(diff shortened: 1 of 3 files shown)


scripts/build.js
```diff
@@ -1,14 +1,19 @@
...
 const { promisify } = require("util");
-const { relative } = require("path");
+const { relative, join } = require("path");
 const copy = promisify(require("copy"));
 const glob = promisify(require("glob"));
 const bytes = require("bytes");
 
 const minify = true;
+const v8cache = true;
+const cache = join(__dirname, "..", ".cache");
```
guybedford, line 11: We shouldn't need to locate a custom cache directory - we can just use the same one used in the default build as that is already versioned.

scripts/build.js
```diff
@@ -1,14 +1,19 @@
...
 
 async function main() {
+  if (process.argv[2] === "--no-cache") {
+    rmdirSync(cache, { recursive: true });
+  }
```
guybedford, line 16: Then `--no-cache` is just `cache: false` I believe.

guybedford (changes requested):
👍

scripts/build.js
styfle: @guybedford What about `v8cache`, should that always be true?

guybedford: That creates a v8 snapshot to speedup the Node.js execution of ncc, so yes.

scripts/build.js
styfle: I like keeping this to the local project since the default ncc cache is a tmp dir like `/var/folders/v7/7kzxdgb81x324tk9pg633s200000gm/T/ncc-cache` so it makes it hard to inspect. But putting it in the project root makes it easier to inspect.

scripts/build.js
```diff
@@ -8,12 +8,9 @@ const bytes = require("bytes");
 
 const minify = true;
 const v8cache = true;
-const cache = join(__dirname, "..", ".cache");
+const cache = process.argv[2] === "--no-cache" ? false : join(__dirname, "..", ".cache");
```
guybedford, line 11: In the cache case, the ncc default is always to cache so we shouldn't need to provide a custom path either.

scripts/build.js
guybedford: I see ok.

guybedford (approved):

PR #669, opened 2021-03-19
Upgrade to terser@5.6.1

guybedford:
This updates to terser@5.6.1 with latest syntax support.

Fixes #661.

PR #703, opened 2021-04-29
fix: handle terser error

guybedford:
With the Terser upgrade in https://github.com/vercel/ncc/pull/669 the error handling needed to be updated to handle the new async form of the function.

As a result Terser errors were becoming build errors, which is fixed here.

This resolves the Terser error regression from https://github.com/vercel/ncc/issues/697.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -462,27 +462,32 @@ function ncc (
     }
 
     if (minify) {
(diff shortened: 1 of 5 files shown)


styfle (approved):
Thanks!

Issue #499, opened 2020-01-16
Private class fields break build


dominiklessel:
I have a project, which uses hapi. With their latest 19.x release they dropped support for node 10 and make use of for example private class fields, which now breaks my ncc build :

```
> ncc build

ncc: Version 0.20.5
ncc: Compiling file index.js
Error: Module parse failed: Unexpected character '#' (17:4)
File was processed with these loaders:
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/empty-loader.js
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/relocate-loader.js
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/shebang-loader.js
You may need an additional loader to handle the result of these loaders.
| exports = module.exports = internals.Auth = class {
| 
>     #core = null;
|     #schemes = {};
|     #strategies = {};
Module parse failed: Unexpected token (51:12)
File was processed with these loaders:
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/empty-loader.js
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/relocate-loader.js
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/shebang-loader.js
You may need an additional loader to handle the result of these loaders.
| exports = module.exports = internals.Core = class {
| 
>     actives = new WeakMap();                                                   // Active requests being processed
|     app = {};
|     auth = new Auth(this);
Module parse failed: Unexpected token (12:9)
File was processed with these loaders:
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/empty-loader.js
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/relocate-loader.js
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/shebang-loader.js
You may need an additional loader to handle the result of these loaders.
| exports = module.exports = internals.Ext = class {
| 
>     type = null;
|     nodes = null;
| 
Module parse failed: Unexpected token (246:10)
File was processed with these loaders:
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/empty-loader.js
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/relocate-loader.js
 * ./node_modules/@zeit/ncc/dist/ncc/loaders/shebang-loader.js
You may need an additional loader to handle the result of these loaders.
| exports.ValidationError = class extends Error {
| 
>     isJoi = true;
|     name = 'ValidationError';
| 
[...]

PR #707, opened 2021-05-09
fix: Update to webpack@5.36.2

adriencohen:
Moving to the latest release of webpack could solve: https://github.com/vercel/ncc/issues/499

The [underlying issue](https://github.com/webpack/webpack/issues/10216) in webpack seems to not totally be closed. but I was able to build a package using private class fields


styfle (approved):
Thanks!

Issue #679, opened 2021-04-01
ES module output mode


guybedford:
With Node.js 10 going EOL soon, it's about time to support a mode for outputting ES modules with ncc, as part of moving towards an ecosystem of published ESM!

PR #720, opened 2021-06-26
major: output ESM for .mjs or "type": "module" builds

guybedford:
This adds support to ncc for a new `esm` option to output ES modules from the build. It defaults to true when the input file is ".mjs" or the input file is contained within a "type": "module" boundary.

The support is based on Webpack's experimental support which seems to have a few edge cases that I have posted in https://github.com/webpack/webpack/issues/13639.

If releasing this feature it would be best to treat it as having an experimental status just like the Webpack feature it exposes.

Resolves https://github.com/vercel/ncc/issues/679.

To be merged after https://github.com/vercel/ncc/pull/723 and https://github.com/vercel/webpack-asset-relocator-loader/pull/140, https://github.com/vercel/ncc/pull/724.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -227,10 +227,24 @@ async function runCmd (argv, stdout, stderr) {
       if (args._.length > 2)
         errTooManyArguments("build");
 
+      function hasTypeModule (path) {
+        let root = pathResolve('/');
+        while ((path = pathResolve(path, '..')) !== root) {
+          try {
+            return JSON.parse(fs.readFileSync(eval('pathResolve')(path, 'package.json')).toString()).type === 'module';
+          }
+          catch (e) {
+            if (e.code === 'ENOENT')
+              continue;
+            throw e;
+          }
+        }
+      }
+
       let startTime = Date.now();
       let ps;
       const buildFile = eval("require.resolve")(resolve(args._[1] || "."));
-      const ext = buildFile.endsWith('.cjs') ? '.cjs' : buildFile.endsWith('.mjs') ? '.mjs' : '.js';
+      const ext = buildFile.endsWith('.cjs') ? '.cjs' : buildFile.endsWith('.mjs') || !hasTypeModule(buildFile) ? '.mjs' : '.js';
       const ncc = require("./index.js")(
         buildFile,
         {
--- a/src/cli.js
+++ b/src/cli.js
@@ -25,7 +25,6 @@ Commands:
 Options:
   -o, --out [file]         Output directory for build (defaults to dist)
   
-  --esm                    Create an ES Module build output
   -m, --minify             Minify output
   -C, --no-cache           Skip build cache population
   -s, --source-map         Generate source map
@@ -131,7 +130,6 @@ async function runCmd (argv, stdout, stderr) {
     args = require("arg")({
       "--debug": Boolean,
       "-d": "--debug",
-      "--esm": Boolean,
       "--external": [String],
       "-e": "--external",
       "--out": String,
@@ -222,7 +220,6 @@ async function runCmd (argv, stdout, stderr) {
       if (existsSync(outDir))
         rimraf.sync(outDir);
       mkdirp.sync(outDir);
-      writeFileSync(eval('resolve')(outDir, 'package.json'), args['--esm'] ? '{ "type": "module" }' : '{}');
       run = true;
 
     // fallthrough
@@ -233,13 +230,12 @@ async function runCmd (argv, stdout, stderr) {
       let startTime = Date.now();
       let ps;
       const buildFile = eval("require.resolve")(resolve(args._[1] || "."));
-      const ext = !args['--esm'] && buildFile.endsWith('.cjs') ? '.cjs' : '.js';
+      const ext = buildFile.endsWith('.cjs') ? '.cjs' : buildFile.endsWith('.mjs') ? '.mjs' : '.js';
       const ncc = require("./index.js")(
         buildFile,
         {
           debugLog: args["--debug"],
           minify: args["--minify"],
-          esm: args["--esm"],
           externals: args["--external"],
           sourceMap: args["--source-map"] || run,
           sourceMapRegister: args["--no-source-map-register"] ? false : undefined,
--- a/src/cli.js
+++ b/src/cli.js
@@ -244,7 +244,7 @@ async function runCmd (argv, stdout, stderr) {
       let startTime = Date.now();
       let ps;
       const buildFile = eval("require.resolve")(resolve(args._[1] || "."));
-      const esm = buildFile.endsWith('.mjs') || buildFile.endsWith('.js') && hasTypeModule(buildFile);
+      const esm = buildFile.endsWith('.mjs') || !buildFile.endsWith('.cjs') && hasTypeModule(buildFile);
       const ext = buildFile.endsWith('.cjs') ? '.cjs' : esm && (buildFile.endsWith('.mjs') || !hasTypeModule(buildFile)) ? '.mjs' : '.js';
       const ncc = require("./index.js")(
         buildFile,
--- a/src/cli.js
+++ b/src/cli.js
@@ -244,7 +244,8 @@ async function runCmd (argv, stdout, stderr) {
       let startTime = Date.now();
       let ps;
       const buildFile = eval("require.resolve")(resolve(args._[1] || "."));
-      const ext = buildFile.endsWith('.cjs') ? '.cjs' : buildFile.endsWith('.mjs') || !hasTypeModule(buildFile) ? '.mjs' : '.js';
+      const esm = buildFile.endsWith('.mjs') || buildFile.endsWith('.js') && hasTypeModule(buildFile);
+      const ext = buildFile.endsWith('.cjs') ? '.cjs' : esm && (buildFile.endsWith('.mjs') || !hasTypeModule(buildFile)) ? '.mjs' : '.js';
       const ncc = require("./index.js")(
         buildFile,
         {
--- a/src/cli.js
+++ b/src/cli.js
@@ -5,7 +5,7 @@ const glob = require("glob");
 const shebangRegEx = require("./utils/shebang");
 const rimraf = require("rimraf");
 const crypto = require("crypto");
-const { writeFileSync, unlink, existsSync, symlinkSync } = require("fs");
+const { writeFileSync, unlink, existsSync, symlinkSync, mkdir } = require("fs");
 const mkdirp = require("mkdirp");
 const { version: nccVersion } = require('../package.json');
 
@@ -24,6 +24,8 @@ Commands:
 
 Options:
   -o, --out [file]         Output directory for build (defaults to dist)
+  
+  --esm                    Create an ES Module build output
   -m, --minify             Minify output
   -C, --no-cache           Skip build cache population
   -s, --source-map         Generate source map
@@ -129,6 +131,7 @@ async function runCmd (argv, stdout, stderr) {
     args = require("arg")({
       "--debug": Boolean,
       "-d": "--debug",
+      "--esm": Boolean,
       "--external": [String],
       "-e": "--external",
       "--out": String,
@@ -218,6 +221,8 @@ async function runCmd (argv, stdout, stderr) {
       );
       if (existsSync(outDir))
         rimraf.sync(outDir);
+      mkdirp.sync(outDir);
+      writeFileSync(eval('resolve')(outDir, 'package.json'), args['--esm'] ? '{ "type": "module" }' : '{}');
       run = true;
 
     // fallthrough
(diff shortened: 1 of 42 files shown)


src/cli.js
```diff
@@ -24,6 +24,8 @@ Commands:
 
 Options:
   -o, --out [file]         Output directory for build (defaults to dist)
+  
+  --esm                    Create an ES Module build output
```
styfle, line 28: I think we can omit this flag and instead detect automatically.

For example:

```js
const esm = buildFile.endsWith('.mjs') ||
  (buildFile.endsWith('.js') && require('package.json').type === 'module')
```

guybedford: It might be a bit magic to do this reverse detection of the package type perhaps? Perhaps we leave the flag as well and treat this as a default though...

styfle: The flag doesn't really make sense in my option because its not really an option. The entrypoint either is or it isn't ESM.

When I run `node index.mjs` then it "just works" so I would expect `ncc run index.mjs` to just work as well.

guybedford: The thing is that the authoring format and the consumption format don't have to be the same. Also we don't currently allow customizing the file name.

Another option could even be to output ES modules by default...

styfle: > the authoring format and the consumption format don't have to be the same

I don't think that is the goal of `ncc`. The number 1 design goal is "Zero configuration" from the README.

We need to follow Node.js conventions here, especially because publish with `npm` will use the default `package.json` which might contain `type: module`.

guybedford: Right now, ncc will convert ES modules into CommonJS for publishing. Changing this default is a breaking change.

styfle: Yep, we can note it as a breaking change

guybedford: Ok, we can do that. Let's try to align it with this webpack PR release then as well - https://github.com/webpack/webpack/pull/13622.

Are there any other breaking changes we should make at the same time? I will aim to update the PR this week.

styfle: I can't think of any other breaking changes we need right now

guybedford: Two possible issues with making the major change initially instead of having an `--esm` flag to start:
1. Updating the self build and all the tests is quite a bit of work
2. The v8cache can't be supported for --esm at all unfortunately.
Let me know if you still want to go ahead with this or use the flag.

guybedford:
I've rebased this PR to the webpack update at https://github.com/vercel/ncc/pull/723.

src/index.js
```diff
@@ -30,14 +30,29 @@ const defaultPermissions = 0o666;
...
 module.exports = ncc;
 function ncc (
   entry,
   {
     cache,
     customEmit = undefined,
+    esm = entry.endsWith('.mjs') || !entry.endsWith('.cjs') && hasTypeModule(entry),
     externals = [],
-    filename = 'index' + (entry.endsWith('.cjs') ? '.cjs' : '.js'),
+    filename = 'index' + (!esm && entry.endsWith('.cjs') ? '.cjs' : !esm || hasTypeModule(entry) ? '.js' : '.mjs'),
```
styfle, line 55: I'm curious why this doesn't follow the same logic as the CLI above on line 233:

```js
const ext = buildFile.endsWith('.cjs') ? '.cjs' : buildFile.endsWith('.mjs') ? '.mjs' : '.js';
```

Seems like we should preserve the file extension here too

src/index.js
```diff
@@ -55,6 +70,9 @@ function ncc (
     production = true,
   } = {}
 ) {
+  if (esm)
+    v8cache = false;
```
styfle, line 74: Let's add a comment why this is false and link to any docs if applicable

src/index.js
```diff
@@ -483,10 +504,10 @@ function ncc (
         
         ({ code, map } = {
           code: result.code,
-          map: sourceMap ? JSON.parse(result.map) : undefined
+          map: map ? JSON.parse(result.map) : undefined
```
styfle, line 507: Do we need to update `map` above so that it includes mjs too?

https://github.com/vercel/ncc/blob/2ba1c2290d8fead03cba66088961ab308fc698f6/src/index.js#L436-L439

src/index.js
guybedford: Good point, I've cleaned the logic up a bit let me know how that looks.

src/index.js
guybedford: The source maps seem to be working to me. Would rather leave this alone as a legacy fix - it may be possible to remove it now possibly.

guybedford: I've posted https://github.com/webpack/webpack/issues/13740 to follow up on this.

guybedford:
There is one last remaining failure here I'm pretty sure is a Webpack bug.

To replicate, clone the repo and run:

```
./src/cli.js build test/fixtures/interop-test.mjs -o bug && node bug/index.mjs
```

This gives the error:

```
const external_assert_namespaceObject = __WEBPACK_EXTERNAL_createRequire(import.meta.url)("assert");
                                        ^

ReferenceError: __WEBPACK_EXTERNAL_createRequire is not defined
```

where the contents of that file are:

```
/******/ var __webpack_modules__ = ({

/***/ 181:
/***/ ((__unused_webpack_module, exports) => {

Object.defineProperty(exports, "__esModule", ({ value: true }));
exports.s = 's';
exports['com' + 'puted'] = 'y';
exports.default = 'z';



/***/ })

/******/ });
/************************************************************************/
/******/ // The module cache
/******/ var __webpack_module_cache__ = {};
/******/ 
/******/ // The require function
/******/ function __nccwpck_require__(moduleId) {
/******/ 	// Check if module is in cache
/******/ 	var cachedModule = __webpack_module_cache__[moduleId];
/******/ 	if (cachedModule !== undefined) {
/******/ 		return cachedModule.exports;
/******/ 	}
/******/ 	// Create a new module (and put it into the cache)
/******/ 	var module = __webpack_module_cache__[moduleId] = {
/******/ 		// no module.id needed
/******/ 		// no module.loaded needed
/******/ 		exports: {}
/******/ 	};
/******/ 
/******/ 	// Execute the module function
/******/ 	var threw = true;
/******/ 	try {
/******/ 		__webpack_modules__[moduleId](module, module.exports, __nccwpck_require__);
/******/ 		threw = false;
/******/ 	} finally {
/******/ 		if(threw) delete __webpack_module_cache__[moduleId];
/******/ 	}
/******/ 
/******/ 	// Return the exports of the module
/******/ 	return module.exports;
/******/ }
/******/ 
/************************************************************************/
/******/ /* webpack/runtime/compat */
/******/ 
/******/ if (typeof __nccwpck_require__ !== 'undefined') __nccwpck_require__.ab = new URL('.', import.meta.url).pathname.slice(import.meta.url.match(/^file:\/\/\/\w:/) ? 1 : 0, -1) + "/";/************************************************************************/
var __webpack_exports__ = {};
// This entry need to be wrapped in an IIFE because it need to be in strict mode.
(() => {
"use strict";

;// CONCATENATED MODULE: external "assert"
const external_assert_namespaceObject = __WEBPACK_EXTERNAL_createRequire(import.meta.url)("assert");
// EXTERNAL MODULE: ./test/fixtures/interop.cjs
var interop = __nccwpck_require__(181);
;// CONCATENATED MODULE: ./test/fixtures/interop-test.mjs



(0,external_assert_namespaceObject.strictEqual)(interop.s, 's');
(0,external_assert_namespaceObject.strictEqual)(/* __esModule */true, true);
(0,external_assert_namespaceObject.strictEqual)(interop.default, 'z');
(0,external_assert_namespaceObject.strictEqual)(interop.s, 's');

})();
```

src/index.js
guybedford: Double-checked and this is definitely only needed for the `.cjs` cases as a webpack hack there. `.mjs` still works fine.

src/index.js
```diff
@@ -288,7 +309,7 @@ function ncc (
     node: false,
     externals ({ context, request }, callback) {
       const external = externalMap.get(request);
-      if (external) return callback(null, `commonjs ${external}`);
+      if (external) return callback(null, `${esm ? 'module' : 'commonjs'} ${external}`);
```
sokra, line 312:
``` js
    externals ({ context, request, dependencyType }, callback) {
      const external = externalMap.get(request);
      if (external) return callback(null, `${dependencyType === 'esm' && esm ? 'module' : 'commonjs'} ${external}`);
```

sokra (changes requested):
Could you try that?

src/index.js
guybedford: Nice I've added this.

guybedford:
Tests are failing here pending webpack bug https://github.com/webpack/webpack/issues/13744.

src/cli.js
```diff
@@ -218,17 +219,33 @@ async function runCmd (argv, stdout, stderr) {
...
 
+      function hasTypeModule (path) {
+        let root = resolve('/');
+        while ((path = resolve(path, '..')) !== root) {
+          try {
+            return JSON.parse(readFileSync(eval('resolve')(path, 'package.json')).toString()).type === 'module';
+          }
+          catch (e) {
+            if (e.code === 'ENOENT')
+              continue;
```
styfle, line 238: Should this be `break` instead of `continue`? I would assume `ENOENT` means that walking up the tree to the root would still yield `ENOENT` but not sure how Node.js handles that case.

src/cli.js
```diff
@@ -218,17 +219,33 @@ async function runCmd (argv, stdout, stderr) {
...
+            throw e;
+          }
+        }
+      }
+
       let startTime = Date.now();
       let ps;
       const buildFile = eval("require.resolve")(resolve(args._[1] || "."));
-      const ext = buildFile.endsWith('.cjs') ? '.cjs' : '.js';
+      const esm = buildFile.endsWith('.mjs') || buildFile.endsWith('.js') && hasTypeModule(buildFile);
```
styfle, line 247: There's a subtle difference in the implementation here and the `esm` default in in `src/index.js` file. Is there a reason for the difference?

src/cli.js
```diff
@@ -218,17 +219,33 @@ async function runCmd (argv, stdout, stderr) {
...
+      const ext = buildFile.endsWith('.cjs') ? '.cjs' : esm && (buildFile.endsWith('.mjs') || !hasTypeModule(buildFile)) ? '.mjs' : '.js';
```
styfle, line 248: I'm curious why you didn't decide to preserve the extension of the input file?

I would expect this:
```
.cjs => .cjs
.mjs => .mjs
.js => .js
```

guybedford: @style we do keep the extension so that `.cjs` -> `.cjs` and `.mjs` -> `.mjs`. The issue with `.js` is that it will be read by webpack as supporting both CJS and ESM, so isn't enough information on its own, hence the boundary check.

src/cli.js
guybedford: This is how the Node.js package boundary lookup works. See https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#packages_package_json_and_file_extensions.

src/cli.js
guybedford: This line of `const esm = ` is the same for both, which lines do you mean?

src/cli.js
styfle: Oh I was thinking of `ENOENT` in the case it means "directory does not exist" but that wouldn't happen since package.json is a file, not a directory 😌

src/cli.js
styfle: They're not the same

## src/cli.js

```js
const esm = buildFile.endsWith('.mjs') || buildFile.endsWith('.js') && hasTypeModule(buildFile);
```

## src/index.js

```js
esm = entry.endsWith('.mjs') || !entry.endsWith('.cjs') && hasTypeModule(entry),
```

## diff

```diff
- esm = entry.endsWith('.mjs') || entry.endsWith('.js') && hasTypeModule(entry)
+ esm = entry.endsWith('.mjs') || !entry.endsWith('.cjs') && hasTypeModule(entry)
```

guybedford: Thanks, yes these should be the same - I've pushed up a fix.

src/cli.js
```diff
@@ -24,6 +24,7 @@ Commands:
 
 Options:
   -o, --out [file]         Output directory for build (defaults to dist)
+  
   -m, --minify             Minify output
```
styfle, line 28:
```suggestion
  -m, --minify             Minify output
```

styfle (approved):
Looks good to me, thanks! 🎉

src/cli.js
```diff
@@ -218,17 +218,33 @@ async function runCmd (argv, stdout, stderr) {
...
       run = true;
 
     // fallthrough
     case "build":
       if (args._.length > 2)
         errTooManyArguments("build");
 
+      function hasTypeModule (path) {
+        let root = resolve('/');
+        while ((path = resolve(path, '..')) !== root) {
```
sokra, line 231: This might break on windows when `root === "C:\"` and `path === "D:\some\path"`, as it never reaches root.

You can do it this way:

```suggestion
        while (path !== (path = resolve(path, '..'))) {
```

src/index.js
```diff
@@ -30,14 +30,29 @@ const defaultPermissions = 0o666;
 
 const relocateLoader = eval('require(__dirname + "/loaders/relocate-loader.js")');
 
+function hasTypeModule (path) {
```
sokra, line 33: This seem to be duplicated code. Can this be moved into a shared file?

sokra (dismissed):

src/cli.js
guybedford: Nice syntax!

guybedford:
Thanks @sokra for the review, updated.

styfle (approved):

src/cli.js
sokra: Usually I would avoid = in conditions, but in this case it's really tense, while moving it out of the condition is really verbose...

PR #723, opened 2021-07-03
Update to webpack@5.42.0

guybedford:
Featuring improved ESM output support per https://github.com/webpack/webpack/pull/13622.

This PR also separates the test suites, done as part of debugging a Node.js 14 regression that was happening on CI when not running against latest Node.js. The CI workflow is updated to always run on the latest of the current Node.js major.


.github/workflows/ci.yml
```diff
@@ -27,6 +27,7 @@ jobs:
       uses: actions/setup-node@v2
       with:
         node-version: ${{ matrix.node }}
+        check-latest: true
```
styfle, line 30: Which version was it using before this option?

styfle: Looks like it bumped from 14.17.1 to 14.17.3 and that fixed a bug?

styfle (approved):
Thanks!

Issue #732, opened 2021-07-22
Error when building ESM project


stayradiated:
First up, just wanted to say how much I love `ncc`, it makes distributing JS code a breeze. Thank you!

I've just upgraded to `v0.29.0`, but when I try to build a project that uses ESM modules (`type: "module"` and `.js` extensions) I get this error:

```
TypeError: Cannot read property 'toString' of undefined
    at finalizeHandler (/project/node_modules/.pnpm/@vercel+ncc@0.29.0/node_modules/@vercel/ncc/dist/ncc/index.js.cache.js:37:1866714)
```

Looking at `./src/index.js` it seems the issue is related to calling `pjsonPath.source.toString()`.
When I log `pgjsonPath` I just get the string `package.json`.

```javascript
      const pjsonPath = (baseDir === '.' ? '' : baseDir) + 'package.json';
      if (assets[pjsonPath])
        assets[pjsonPath].source = JSON.stringify(Object.assign(JSON.parse(pjsonPath.source.toString()), { type: 'module' }));
```

https://github.com/vercel/ncc/blob/main/src/index.js#L532-L537

guybedford:
Thanks for the clear report and glad to hear the project is working well for you otherwise.

This should be resolved in https://github.com/vercel/ncc/pull/733, may still take a few hours for the merge and release.

PR #733, opened 2021-07-22
fix: package.json asset type module setting

guybedford:
This fixes https://github.com/vercel/ncc/issues/732, which was a feature added to ensure any emitted `package.json` doesn't stop the `"type": "module"` boundary from applying, but had a typo in the reading process.


styfle (approved):
Nice catch, thanks!

PR #751, opened 2021-08-19
fix: ensure nested builds of __nccwpck_require__

guybedford:
This fixes the `__nccwpck_require__` handling to ensure conflicts are avoided when running against a build that already has an `__nccpck_require__` identifier. It uses numbering up to 9 given the limited space. We can look at increasing this limit in future though.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -546,8 +546,22 @@ function ncc (
     // __webpack_require__ can conflict with webpack injections in module scopes
     // to avoid this without recomputing the source map we replace it with an
     // identical length identifier
-    if (code.indexOf('"__webpack_require__"') === -1)
+    if (code.indexOf('"__webpack_require__"') === -1) {
+      // dedupe any existing __nccwpck_require__ first
+      if (code.indexOf('__nccwpck_require2_') !== -1) {
+        // nth level nesting (we support 9 levels apparently)
+        for (let i = 9; i > 1; i--) {
+          if (code.indexOf(`__nccwpck_require${i}_`) === -1)
+            continue;
+          if (i === 9)
+            throw new Error('9 levels of ncc build nesting reached, please post an issue to support this level of ncc build composition.');
+          code = code.replace(/__nccwpck_require${i}_/g, `__nccwpck_require${i + 1}_`);
+        }
+      }
+      if (code.indexOf('__nccwpck_require__') !== -1)
+        code = code.replace(/__nccwpck_require__/g, '__nccwpck_require2_');
       code = code.replace(/__webpack_require__/g, '__nccwpck_require__');
+    }
 
     // for each .js / .mjs / .cjs file in the asset list, build that file with ncc itself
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -546,8 +546,22 @@ function ncc (
...
+          if (code.indexOf(`__nccwpck_require${i}_`) === -1)
+            continue;
+          if (i === 9)
+            throw new Error('9 levels of ncc build nesting reached, please post an issue to support this level of ncc build composition.');
+          code = code.replace(/__nccwpck_require${i}_/g, `__nccwpck_require${i + 1}_`);
+        }
+      }
+      if (code.indexOf('__nccwpck_require__') !== -1)
+        code = code.replace(/__nccwpck_require__/g, '__nccwpck_require2_');
       code = code.replace(/__webpack_require__/g, '__nccwpck_require__');
```
styfle, line 563: Do we need to replace all 9 levels here too?

src/index.js
```diff
@@ -546,8 +546,22 @@ function ncc (
...
+    if (code.indexOf('"__webpack_require__"') === -1) {
+      // dedupe any existing __nccwpck_require__ first
+      if (code.indexOf('__nccwpck_require2_') !== -1) {
+        // nth level nesting (we support 9 levels apparently)
+        for (let i = 9; i > 1; i--) {
+          if (code.indexOf(`__nccwpck_require${i}_`) === -1)
+            continue;
+          if (i === 9)
+            throw new Error('9 levels of ncc build nesting reached, please post an issue to support this level of ncc build composition.');
+          code = code.replace(/__nccwpck_require${i}_/g, `__nccwpck_require${i + 1}_`);
```
styfle, line 558: Are you sure this is correct? The regex `/__nccwpck_require${i}_/g` seems like its supposed to use the value of `i` but its not currently used.

src/index.js
```diff
@@ -546,8 +546,22 @@ function ncc (
     // __webpack_require__ can conflict with webpack injections in module scopes
     // to avoid this without recomputing the source map we replace it with an
     // identical length identifier
-    if (code.indexOf('"__webpack_require__"') === -1)
+    if (code.indexOf('"__webpack_require__"') === -1) {
+      // dedupe any existing __nccwpck_require__ first
+      if (code.indexOf('__nccwpck_require2_') !== -1) {
+        // nth level nesting (we support 9 levels apparently)
+        for (let i = 9; i > 1; i--) {
+          if (code.indexOf(`__nccwpck_require${i}_`) === -1)
```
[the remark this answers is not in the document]
guybedford: No, we go straight to 2.

src/index.js
guybedford: Excellent catch thank you - I've fixed this up and added a test for the subsequent numbering levels of nesting.

styfle (approved):
Thanks!

PR #766, opened 2021-09-09
Fix `compilerOptions` from `tsconfig.json`

### LINKED ISSUE #765: Error: ES Modules may not assign module.exports or exports.*, Use ESM export syntax
When using typescript with `"module": "commonjs"` in `tsconfig.json`, mixing ESM/CJS fails with the following error:

```
Error: ES Modules may not assign module.exports or exports.*, Use ESM export syntax
```

This does not match the behavior of `tsc` which allows mixed usage.

```
git clone https://github.com/uncurated-tests/ncc-ts-mixed-modules
cd ncc-ts-mixed-modules
yarn install

yarn run tsc
node dist/index.js # works

rm -rf dist

ncc build index.ts
node dist/index.js # fails
```

https://github.com/uncurated-tests/ncc-ts-mixed-modules

styfle:
This will make sure we pass `tsconfig.json` to the loader.

In particular the `module: commonjs` option is important to ensure mixed module usage like `tsc` supports.

Fixes #765

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -109,7 +109,7 @@ function ncc (
   // add TsconfigPathsPlugin to support `paths` resolution in tsconfig
   // we need to catch here because the plugin will
   // error if there's no tsconfig in the working directory
-  let fullTsconfig;
+  let fullTsconfig = {};
   try {
     const tsconfig = tsconfigPaths.loadConfig();
     fullTsconfig = loadTsconfig(tsconfig.configFileAbsolutePath) || {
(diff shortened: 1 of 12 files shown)


src/index.js
```diff
@@ -335,14 +335,10 @@ function ncc (
               compiler: eval('__dirname + "/typescript.js"'),
               compilerOptions: {
                 allowSyntheticDefaultImports: true,
-                module: 'esnext',
-                outDir: '//',
-                ...(fullTsconfig &&
-                  fullTsconfig.compilerOptions &&
-                  fullTsconfig.compilerOptions.incremental
-                    ? { incremental: false }
-                    : {}),
```
guybedford, line 344: Wait I only just saw this isn't using the TSConfig at all. Shouldn't this be `...fullTsconfig.compilerOptions` for all compiler options?

styfle: Yeah good point, we'll see if all the tests pass or if there are other options we need to override

src/index.js
```diff
@@ -334,9 +334,8 @@ function ncc (
               transpileOnly,
               compiler: eval('__dirname + "/typescript.js"'),
               compilerOptions: {
+                ...fullTsconfig.compilerOptions,
```
guybedford, line 337: We should still put `module: 'esnext'` as the default, before this line so that it gets overridden:

```suggestion
                module: 'esnext'
                ...fullTsconfig.compilerOptions,
```

TooTallNate (approved):

PR #770, opened 2021-09-13
Fix `tsconfig.json` detection

styfle:
This PR fixes a bug seen in PR https://github.com/vercel/vercel/pull/6700 (see [logs](https://github.com/vercel/vercel/pull/6700/checks?check_run_id=3562974831#step:12:226)) where the `tsconfig.json` is in a parent directory.

The `tsconfig-paths` package was returning an error [`Missing baseUrl in compilerOptions`](https://github.com/dividab/tsconfig-paths/blob/0b259d4cf6cffbc03ad362cfc6bb129d040375b7/src/config-loader.ts#L80-L85).

So instead, we'll manually walk up the filesystem to find the relevant `tsconfig.json` like we do in `@vercel/build-utils`.

### DIFF
--- a/test/unit/ts-target-es2018/output-coverage.js
+++ b/test/unit/ts-target-es2018/output-coverage.js
@@ -1,0 +1,16 @@
+/******/ (() => { // webpackBootstrap
+/******/ 	"use strict";
+/******/ 	/* webpack/runtime/compat */
+/******/ 	
+/******/ 	if (typeof __nccwpck_require__ !== 'undefined') __nccwpck_require__.ab = __dirname + "/";
+/******/ 	
+/************************************************************************/
+var __webpack_exports__ = {};
+
+if (process === null || process === void 0 ? void 0 : process.env.FOO) {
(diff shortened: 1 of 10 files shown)


test/unit/ts-target-es2018/output-coverage.js
```diff
@@ -0,0 +1,16 @@
...
+/******/ 	/* webpack/runtime/compat */
+/******/ 	
+/******/ 	if (typeof __nccwpck_require__ !== 'undefined') __nccwpck_require__.ab = __dirname + "/";
+/******/ 	
+/************************************************************************/
+var __webpack_exports__ = {};
+
+if (process === null || process === void 0 ? void 0 : process.env.FOO) {
+    console.log('foo');
+}
```
styfle, line 12: @TooTallNate Here we can see the optional chaining operator is properly down-leveled 😌

mrmckeb (approved):
It might be worth extracting `walkParentDirs` out, but not sure _where_...

PR #773, opened 2021-09-13
fix: Add missing variable declaration

styfle:
I accidentally removed a variable in PR #770 

We should convert ncc source code to TS to catch these types of errors.

_Update_: this variable seems unused 🤷‍♂️ 

- `tsconfigMatchPath` added in #151
- `tsconfigMatchPath` removed in #378


TooTallNate (approved):

Timer (approved):
LGTM if you wanna ship, otherwise, 📕! Maybe the side effect was important?

PR #775, opened 2021-09-16
Use custom ts paths resolve plugin

styfle:


src/index.js
```diff
@@ -105,31 +104,27 @@ function ncc (
...
 
-    const tsconfigPathsOptions = { silent: true }
-    if (fullTsconfig.compilerOptions.allowJs) {
-      tsconfigPathsOptions.extensions = SUPPORTED_EXTENSIONS
-    }
-    resolvePlugins.push(new TsconfigPathsPlugin(tsconfigPathsOptions));
+  let tsconfig = {};
+  try {
+    const configPath = join(process.cwd(), 'tsconfig.json');
+    const contents = fs.readFileSync(configPath, 'utf8')
```
huozhi, line 111: How about using ts api to parse the config? It has type hint

```
const config = ts.readConfigFile(tsconfigPath, ts.sys.readFile).config;
const compilerOptions = ts.parseJsonConfigFileContent(
        config,
        ts.sys,
        './',
).options;
```

styfle: Sounds good. I'll probably go with the other PR instead https://github.com/vercel/ncc/pull/777

PR #777, opened 2021-09-17
Use swc loader for typescript compilation

styfle:


styfle:
@guybedford @sokra @kdy1 @huozhi Any thoughts on how to solve this one?

Looks like `module.exports = require("shebang-loader");`

is being converted into `module.exports = eval("require")("shebang-loader");`

But I would expect the contents of the `shebang-loader` package instead.

kdy1:
It seems like node failed to resolve `@swc/core-darwin-x64` from `./dist/ncc/loaders/swc-loader.js`.

But I'm not sure about the reason, as I found `yarn install` in the CI script.

src/index.js
```diff
@@ -333,17 +324,30 @@ function ncc (
...
+                strictMode: true,
+                lazy: false,
+                noInterop: !compilerOptions.esModuleInterop
+              },
+              jsc: {
+                externalHelpers: false,
+                keepClassNames: true,
+                target: compilerOptions.target && compilerOptions.target.toLowerCase() || 'es2021',
+                paths: compilerOptions.paths,
+                baseUrl: compilerOptions.baseUrl,
```
styfle, line 344: @kdy1 The `paths` and `baseUrl` are failing with:

```
panicked at 'failed to get current directory: Error { kind: Unsupported, message: "operation not supported on this platform" }', /home/runner/work/swc/swc/ecmascript/transforms/module/src/path.rs:81:56

Stack:

Error: 
    at u.exports.__wbg_new_59cb74e423758ede (/Users/styfle/Code/vercel/ncc/dist/ncc/loaders/swc-loader.js.cache.js:1:2692)
    at null.<anonymous> (wasm://wasm/04480186:0:15193164)
    at null.<anonymous> (wasm://wasm/04480186:0:16752724)
    at null.<anonymous> (wasm://wasm/04480186:0:16566604)
    at null.<anonymous> (wasm://wasm/04480186:0:16677644)
    at null.<anonymous> (wasm://wasm/04480186:0:16749249)
    at null.<anonymous> (wasm://wasm/04480186:0:16736138)
    at null.<anonymous> (wasm://wasm/04480186:0:16687572)
    at null.<anonymous> (wasm://wasm/04480186:0:5322716)
    at null.<anonymous> (wasm://wasm/04480186:0:15233495)
```

styfle: @kdy1 The new `@swc/wasm` no longer panics, but it still fails with very strange behavior.

Input:
```
import module from "@module";

console.log(module);
```

Output:
```
/******/ 	var __webpack_modules__ = ({

/***/ 105:
/***/ ((module) => {

module.exports = eval("require")("../../../../../../../../.././module");


/***/ })

/******/ 	});
```

Results in
```
Error: Cannot find module '../../../../../../../../.././module'
```

harisvsulaiman:
Now that swc is extensively used at vercel, swc support in ncc would be hugely beneficial.

PR #780, opened 2021-09-21
Remove all comments during minification

silverwind:
By default, terser retains "some" comments during minifaction. Remove all comments instead to reduce output file size.

Ref: https://github.com/terser/terser/blob/master/README.md#format-options


src/index.js
```diff
@@ -480,6 +480,9 @@ function ncc (
             keep_classnames: true,
             keep_fnames: true
           },
+          format: {
+            comments: false
+          },
```
styfle, line 485: I'm not sure we want to do this. The docs say:

> by default it keeps JSDoc-style comments that contain "@license", "@preserve" or start with `!`

Some code relies on comments (I think Angular does func.toString() for example) so if someone wanted it to be preserved we should probably preserve it.

silverwind: While I think code that relies on `function.toString` is fundamentally flawed, I guess a option could be introduced on whether to retain such comments.

Or more maybe even more generally, the minifier options could be exposed on the [`Node.js API`](https://github.com/vercel/ncc#programmatically-from-nodejs), e.g. property `minfierOptions` that is then deep-merged into the static configuration.

styfle: I would rather not expose terser options since we might replace it with something else.

Are you seeing big differences in bundle sizes with this change?

I noticed you didn't add a test.

Timer (changes requested):
I don't believe we can exclude license comments by default, this PR would have to be a larger scope of work that extracts license info to a different file.

src/index.js
silverwind: It saves a few kilobytes, but it's nothing major in most cases.

silverwind:
I'm fine if you don't want to have it. I'm using `perl -0777 -p -i -e 's#\n?\/\*![\s\S]*?\*\/\n?##g' bundle.js` to remove all comments from the bundle now.

PR #790, opened 2021-10-08
add `jsonc-loader`

beanjs:
[ x ]  Test


src/index.js
```diff
@@ -17,7 +17,7 @@ const { hasTypeModule } = require('./utils/has-type-module');
 // support glob graceful-fs
 fs.gracefulify(require("fs"));
 
-const SUPPORTED_EXTENSIONS = [".js", ".json", ".node", ".mjs", ".ts", ".tsx"];
+const SUPPORTED_EXTENSIONS = [".js", ".json",".jsonc", ".node", ".mjs", ".ts", ".tsx"];
```
styfle, line 20: Does Node.js support jsonc files?

If not, then this seems to be out of scope for ncc.

PR #801, opened 2021-10-26
chore: Add Node.js 16 to CI

styfle:
https://github.com/nodejs/node/releases/tag/v16.13.0


styfle:
I removed a couple dependencies that weren't working:

- [highlights](https://npmjs.com/package/highlights) which only has 500 downloads each week
- [pdf2json](https://www.npmjs.com/package/pdf2json) which has 50,000 downloads each week 

We'll need to revisit if this is an actual bug in `ncc`

Issue #805, opened 2021-10-31
Fails with Node.js 17


jesec:
Bug: webpack/webpack#14532

styfle:
Fixed in #809

jesec:
For some reasons, #809 and 0.32.0 does not resolve the issue.

```
ncc: Version 0.32.0
ncc: Compiling file index.js into CJS
Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:67:19)
    at Object.createHash (node:crypto:130:10)
    at hashOf (/home/jc/flood/node_modules/@vercel/ncc/dist/ncc/index.js.cache.js:37:1890023)
    at ncc (/home/jc/flood/node_modules/@vercel/ncc/dist/ncc/index.js.cache.js:37:1894488)
    at runCmd (/home/jc/flood/node_modules/@vercel/ncc/dist/ncc/cli.js.cache.js:1:52268)
    at Object.819 (/home/jc/flood/node_modules/@vercel/ncc/dist/ncc/cli.js.cache.js:1:48838)
    at __webpack_require__ (/home/jc/flood/node_modules/@vercel/ncc/dist/ncc/cli.js.cache.js:1:59076)
    at /home/jc/flood/node_modules/@vercel/ncc/dist/ncc/cli.js.cache.js:1:59288
    at /home/jc/flood/node_modules/@vercel/ncc/dist/ncc/cli.js.cache.js:1:59349
    at Object.<anonymous> (/home/jc/flood/node_modules/@vercel/ncc/dist/ncc/cli.js:8:28) {
  opensslErrorStack: [ 'error:03000086:digital envelope routines::initialization error' ],
  library: 'digital envelope routines',
  reason: 'unsupported',
  code: 'ERR_OSSL_EVP_UNSUPPORTED'
}
```

jesec:
The issue may have something to do with ncc's failure to bundle the WebAssembly modules. I encounter some issues with the WebAssembly modules in the past. However, in this case, the issue could prevent webpack from using its own hash implementation:

https://github.com/webpack/webpack/tree/main/assembly/hash

This could become a bigger problem in next version of webpack, since xxhash64 would become default, and there is no OpenSSL implementation to fallback to.

styfle:
Can you share steps to reproduce the issue? Then @sokra can take a look too.

jesec:
It is very simple. 

Install Node v17, and run ncc on any project.

styfle:
[...]

PR #809, opened 2021-11-05
feat: bump to webpack@5.61.0

jesec:
Bug: #805, webpack/webpack#14532

### DIFF
--- a/test/watcher.test.js
+++ b/test/watcher.test.js
@@ -71,7 +71,7 @@ class CustomWatchFileSystem {
     this.paused = false;
     this.changeCallback = changeCallback;
 
-    // ...Start watching files, dirs, mising
+    // ...Start watching files, dirs, missing
     setImmediate(() => {
       this.watchStart(files, dirs, missing);
     });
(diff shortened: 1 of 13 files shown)


styfle (approved):
Thanks!

test/watcher.test.js
```diff
@@ -101,7 +101,6 @@ it('Should support custom watch API', async () => {
   await new Promise((resolve, reject) => {
     const watcher = new CustomWatchFileSystem(function watchStart (files, dirs, missing) {
       expect(files._set.size).toBeGreaterThan(100);
-      expect(missing._set.size).toBeGreaterThan(40);
```
styfle, line 104: Why was `missing` removed here?

jesec: I found it unlikely to be useful.

As seen in earlier commit, https://github.com/vercel/ncc/commit/1bd2aac221b2e53ce2fc0a7c7f79425518310526#diff-b886328ab4b91304e8f0ff17323f05d868c3f83cbb0a6c492f362fdedcd70f96, the line was changed to accommodate the new version of webpack.

styfle: @guybedford Thoughts on this? I'm fine with removing it.

PR #818, opened 2021-11-22
fix: Fix source maps

fenix20113:
* Fix missing source map register if the v8 cache enabled
* Fix missing source maps in assets if the v8 cache disabled


styfle (commented):
Thanks for the PR!

It would be great to add a test to show this fixes the intended bug 👍

fenix20113:
Hi @styfle , I updated the PR with tests

styfle (approved):
Great work, thanks! 🎉

PR #825, opened 2021-12-01
feat: Allow using matches from externals for regex matching

ijjk:
This allows using matches from regex externals to allow maintaining exact require paths without having to list every exact require in the externals config. This allows fixing a longstanding issue in Next.js with ensuring `caniuse-lite` data is externalized properly since `autoprefixer` uses exact requires which can be seen [here](https://unpkg.com/autoprefixer@10.4.0/data/prefixes.js). Related TODO in Next.js [here](https://github.com/vercel/next.js/blob/f2251793782e272e5a74f97bd4f32c94fc705769/packages/next/taskfile.js#L23).

This data being incorrectly bundled with ncc can be seen in this PR https://github.com/vercel/next.js/pull/31984


styfle (approved):
Looks good, thanks!

PR #868, opened 2022-02-04
fix: use sha instead of deprecated md5 for hash algorithm

### LINKED ISSUE #805: Fails with Node.js 17

shakefu:
OpenSSL deprecates and fully removes "md5" as a hash algorithm. This pull request replaces references to "md5" with "sha".

This should resolve the `ERR_OSSL_EVP_UNSUPPORTED` errors.

Should theoretically fix #805 but I am unable to get tests to pass locally, which seems unrelated.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -218,7 +218,7 @@ async function runCmd (argv, stdout, stderr) {
 
       outDir = resolve(
         require("os").tmpdir(),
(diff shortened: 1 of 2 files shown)


styfle (approved):
Thanks!

markoenix (commented):
'md2' in 221

Issue #593, opened 2020-09-22
Bug when compiling with Knex.js on windows 10


code-ape:
We've run into a strange issue with code compilation by ncc resulting in a crashing program on Windows 10. A repo has been created that has been verified to exhibit this issue, here: https://github.com/urbdyn/ncc-bug-1

The bug results in code within Knex.js replacing string matches on imports with full paths. The repo has great amounts of detail and compiled output from a Windows and MacOS machine for comparison.

The resulting compiled code in question that seems to be resulting for this look like this on MacOS (works fine, non-buggy):

```js
/***/ 71330:
/***/ ((module, __unused_webpack_exports, __webpack_require__) => {

function __ncc_wildcard$0 (arg) {
  if (arg === "mssql") return __webpack_require__(19427);
  else if (arg === "mysql") return __webpack_require__(75941);
  else if (arg === "mysql2") return __webpack_require__(17653);
  else if (arg === "oracle") return __webpack_require__(9051);
  else if (arg === "oracledb") return __webpack_require__(25310);
  else if (arg === "postgres") return __webpack_require__(64482);
  else if (arg === "redshift") return __webpack_require__(97332);
  else if (arg === "sqlite3") return __webpack_require__(85370);
}
```

but for some reason is transformed into the following in a Windows environment:

```js
/***/ 71330:
/***/ ((module, __unused_webpack_exports, __webpack_require__) => {

function __ncc_wildcard$0 (arg) {
  if (arg === "C:/code/urbdyn/ncc-bug-1/node_modules/knex/lib/dialects/") return __webpack_require__(19427);
  else if (arg === "C:/code/urbdyn/ncc-bug-1/node_modules/knex/lib/dialects/") return __webpack_require__(75941);
  else if (arg === "C:/code/urbdyn/ncc-bug-1/node_modules/knex/lib/dialects/") return __webpack_require__(17653);
  else if (arg === "C:/code/urbdyn/ncc-bug-1/node_modules/knex/lib/dialects/") return __webpack_require__(9051);
  else if (arg === "C:/code/urbdyn/ncc-bug-1/node_modules/knex/lib/dialects/") return __webpack_require__(25310);
  else if (arg === "C:/code/urbdyn/ncc-bug-1/node_modules/knex/lib/dialects/") return __webpack_require__(64482);
  else if (arg === "C:/code/urbdyn/ncc-bug-1/node_modules/knex/lib/dialects/") return __webpack_require__(97332);
[...]

PR #896, opened 2022-03-25
chore: add windows to CI

styfle:
- Related to https://github.com/vercel/webpack-asset-relocator-loader/pull/155
- Related to https://github.com/vercel/ncc/issues/593

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
         node: [14, 16]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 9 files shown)


styfle:
@yichi-yang Any thoughts what could be wrong here? Looks like the tests already strip `\r` but its still showing up

https://github.com/vercel/ncc/blob/905222e371dbf7f04c3b6dc39660b9dc58da4933/test/unit.test.js#L22-L23

yichi-yang:
> @yichi-yang Any thoughts what could be wrong here? Looks like the tests already strip `\r` but its still showing up

I'm not super familiar with the unit test setup here, but just by looking at the code my guess would be it's `actualSourceMap` that contain `\r`s?

https://github.com/vercel/ncc/blob/905222e371dbf7f04c3b6dc39660b9dc58da4933/test/unit.test.js#L81

styfle:
That one also has a replace, so still not sure why it fails CI 🤔 

https://github.com/vercel/ncc/blob/905222e371dbf7f04c3b6dc39660b9dc58da4933/test/unit.test.js#L76-L79

yichi-yang:
Ah I think you need `/\\r/g` instead of `/\r/g`. Those `\r`s in map appear to already have been escaped.

PR #897, opened 2022-03-26
fix: Disable compile time evaluation of import.meta.url

bjoluc:
I'm affected by https://github.com/webpack/webpack/issues/14445, fixed in https://github.com/webpack/webpack/pull/15246, available in webpack >= [v5.68.0](https://github.com/webpack/webpack/releases/tag/v5.68.0), so I'd love to have an ncc version using webpack >= v5.68.0!

EDIT: Forgot to actually set the corresponding config option at first, added that in f19ab9c.


styfle (commented):
Thanks for the PR!

Can you add a test to ensure this behavior is working as intended?

src/index.js
```diff
@@ -376,7 +376,10 @@ function ncc (
             loader: eval('__dirname + "/loaders/shebang-loader.js"')
           }]
         }
-      ]
+      ],
+      parser: {
+        javascript: { importMeta: false },
+      },
```
styfle, line 382: Since ncc supports both CJS and ESM modes, I think we need to make this conditional based on the entrypoint.

bjoluc: Right, this certainly isn't necessary in CJS modules (though I guess leaving it in the config then wouldn't matter). Thinking about this, what if an ESM dependency of a CJS module uses `import.meta`? I guess the current config would also affect the ESM module. Anyway, I'm about to add some tests for this!

styfle: Hmm good point! I tried `console.log(import.meta.url)` in a cjs file and node gave me an error

> SyntaxError: Cannot use 'import.meta' outside a module

So should be fine to always enable, as long as this is the only behavior thats changed.

Speaking of which, I can't seem to find  any documentation on `importMeta`. Did you mean `importMetaContext`?

https://webpack.js.org/configuration/module/

bjoluc: > Speaking of which, I can't seem to find any documentation on importMeta. Did you mean importMetaContext?

`importMeta` is just right – it looks like it didn't yet make it into the docs; it's in the config types though.

styfle:
Looks like the issue is here https://github.com/vercel/ncc/blob/a9c07e0198abace65147b2bb8962c18297cf8e93/test/watcher.test.js#L79-L92

The returned object no longer matches the latest webpack expected type here https://github.com/webpack/webpack/blob/d3a0f8de03f26a83b4d5db3cfe177617a3801df3/types.d.ts#L12062-L12097

styfle:
Looks like the issue is here https://github.com/vercel/ncc/blob/a9c07e0198abace65147b2bb8962c18297cf8e93/test/watcher.test.js#L79-L92

The returned object no longer matches the latest webpack expected type here https://github.com/webpack/webpack/blob/d3a0f8de03f26a83b4d5db3cfe177617a3801df3/types.d.ts#L12062-L12097

src/index.js
aruniverse: there now, https://webpack.js.org/configuration/module/#moduleparserjavascriptimportmeta

aruniverse:
@bjoluc @styfle anyway to help push this pr along?

bjoluc:
@styfle Thanks for pointing me to the watcher implementation!

I updated the `CustomWatchFileSystem` class according to the new one [over at webpack](https://github.com/webpack/webpack/blob/main/lib/node/NodeWatchFileSystem.js) and `watcher.test.js` passes again. I didn't implement the bunch of deprecated getters from the `Watcher` interface in favor of the generic `getInfo()` which seems to be enough. Not an expert here though. Maybe @guybedford can quickly skim over a4ce744 and check if anything looks wildly wrong?

yarn.lock
```diff
@@ -2313,10 +2313,10 @@
   dependencies:
     "@types/ms" "*"
 
-"@types/eslint-scope@^3.7.0":
-  version "3.7.0"
-  resolved "https://registry.yarnpkg.com/@types/eslint-scope/-/eslint-scope-3.7.0.tgz#4792816e31119ebd506902a482caec4951fabd86"
-  integrity sha512-O/ql2+rrCUe2W2rs7wMR+GqPRcgB6UiqN5RhrR5xruFlY7l9YLMn0ZkDzjoHLeiFkR8MCQZVudUuuvQ2BLC9Qw==
+"@types/eslint-scope@^3.7.3":
+  version "3.7.3"
+  resolved "https://registry.npmjs.org/@types/eslint-scope/-/eslint-scope-3.7.3.tgz#125b88504b61e3c8bc6f870882003253005c3224"
```
styfle, line 2318: This should use yarnpkg.com

bjoluc: Sorry!

styfle (approved):
Great work, thanks! 🎉

PR #926, opened 2022-05-11
major: replace `ts-loader` with swc

kwonoj:
## Breaking Changes

- No more type checking, run `tsc` instead
- No more emit of `.d.ts` files, run `tsc` instead

## Description

PR reattempt #777 with updated dependencies, as well as fixing few immediate code paths to pass existing tests. There are few tests removed, related with ts-error fixtures which is redundant now as swc doesn't run type check while compiling.

Some of fixtures has changed its output mostly due to differences between how swc / tsc emits transformed data. I read it manually and looks like it's non-breaking changes, but still need some eyes.

- Closes #777


styfle:
Thanks for the PR!

I've testing on real repos here:

- https://github.com/vercel/next.js/pull/37200
- https://github.com/vercel/vercel/pull/7875

This needs further investigation before we land since its not clear why the `require()` in TS behaves differently

kwonoj:
> its not clear why the require() in TS behaves differently

Agreed, somehow only on swc it try to replace it into bundled require logic. I may need further digging to find out reason.

styfle:
@kwonoj I found a bug where the `.d.ts` files are not emitted with ncc even thought the tsconfig.json has `"declaration": true`. You can see this had to be changed in `vercel/vercel` [af779ab](https://github.com/vercel/vercel/pull/7875/commits/af779ab72024608cb54c4667850d30cfd0d939ba)

kwonoj:
> even thought the tsconfig.json has "declaration": true

SWC does not have capability to emit type declaration, this would be (somewhat) expected behavior in this changes, since PR entirely replaces typescript into SWC.

Should we preserve type declaration emission feature? if that's the case, sounds like that's a hard blocker to push SWC for ncc.

styfle:
@kwonoj Hmm, I think replacing tsloader with swc is a fair tradeoff to losing `.d.ts` since swc is 2x faster in `vercel/vercel` and `vercel/next.js`. In the cases we need `.d.ts`, we can run the slower `tsc` to generate.

Please update this PR to fix the conflicts and then we can merge, thanks!

test/cli.js
```diff
@@ -60,16 +60,6 @@ module.exports = [
...
   },
-  {
-    args: ["run", "test/fixtures/ts-error2/ts-error.ts"],
-    expect (code, stdout, stderr) {
-      return code === 1 && stderr.toString().indexOf('ts-error.ts(3,16)') !== -1 && stderr.toString().split('\n').length < 11;
-    }
-  },
-  {
-    args: ["run", "-t", "test/fixtures/with-type-errors/ts-error.ts"],
-    expect: { code: 0 }
```
styfle, line 71: Why were these 3 tests removed?

kwonoj: These tests are specific to verify tsc to fail if given typescript file contains error: for example, test verifies output 

```
code === 1 && stderr.toString().indexOf('ts-error.ts(3,16)') !== -1
```

to check tsc's error message contains error line to 3:16 where it have type-related error:

```
//3:16 points 'Y', since there's no corresponding types
function p (x: Y) { 
```

which doesn't have any meaningful behavior in swc as it strips out typescript and does not emit any type check failures.

SukkaW:
> > its not clear why the require() in TS behaves differently
> 
> Agreed, somehow only on swc it try to replace it into bundled require logic. I may need further digging to find out reason.

@kwonoj @styfle 

FYI, the following code behaves differently in tsc and swc:

```
// a.js
console.log('a')

// c.js
console.log('c')

// main.js
import a from './a'
console.log('b')
import b from './c'
```

To be fully compliant with the ESM spec, `main.js` should yield `a c b`, but with tsc it will yield `a b c`.

I don't know if it is the case, so I will raise it here.

kwonoj:
> To be fully compliant with the ESM spec, main.js should yield a c b, but with tsc it will yield a b c.

This is true, but afaik the issue in here I observed was not this one.

test/cli.js
styfle: Ok so to be clear, this PR introduces another breaking change: no more type checking?

kwonoj: Yes, it removes out of the box support from running ncc. If we are considering that is a mandatory feature we need to support, I'm afraid it's not achievable via swc.

styfle: Ok I update the PR title and description to mention this is semver major and has breaking changes.

Also, it seems there are still failing tests

kwonoj: Huh, weird that only subset of node.js 16 CI's failing. I'll try to peek.

kwonoj:
@styfle Looks like CI failures are flaky, need additional investigation.

Currently PR passes all the jobs, but the last commit is dummy (https://github.com/vercel/ncc/pull/926/commits/fe3837b46a8be6f471426488791fa327b3ff5673) which only increases timeout + console.out for stdout/err compare to previous commit failed on some jobs.

styfle (dismissed):

src/index.js
```diff
@@ -108,32 +106,24 @@ function ncc (
...
   }
+
+  let tsconfig = {};
   const resolvePlugins = [];
-  // add TsconfigPathsPlugin to support `paths` resolution in tsconfig
-  // we need to catch here because the plugin will
-  // error if there's no tsconfig in the working directory
-  let fullTsconfig = {};
+  const resolveModules = [];
+  const compilerOptions = tsconfig.compilerOptions || {};
```
SukkaW, line 113: What does this line do? Why assign `compilerOptions` before `tsconfig.json` is even read?

cb1kenobi:
I'm not sure what's going on, but I'm seeing a huge performance regression. I have a super small test package with a `src/index.ts` (contains only 1 exported function) and `src/types.ts` (contains 3 tiny interfaces). My package is an ESM package: `"type": "module"`.

The latest `ncc` v0.34.0 can consistently build my package in about 1.5-2 seconds. Command is `ncc build src/index.ts --minify`. I get type checking and type definition files.

With this PR, it consistently takes 7-12 seconds. Command is identical to above: `ncc build src/index.ts --minify`. I don't get type checking or type definition files.

Setting the `--target es2020` has no effect on the performance. Omitting `--minify` does shave off around 1 second.

My test set up is WSL (Ubuntu) on Windows 11, Node 16.14.2.

On the plus side, the PR produces the exact same output as the latest `ncc`, so my package works as expected.

src/index.js
kwonoj: probably straightforward mistake? 😓

kwonoj:
> performance regression

The only thing I can think of is warmup time to compile wasm modules (`@swc/wasm`), but that probably won't take those huge time unless we spawn bunch of different process and each of them compiles wasm binary.

src/index.js
```diff
@@ -354,17 +344,30 @@ function ncc (
...
-              transpileOnly,
-              compiler: eval('__dirname + "/typescript.js"'),
-              compilerOptions: {
-                module: 'esnext',
-                target: 'esnext',
-                ...fullTsconfig.compilerOptions,
-                allowSyntheticDefaultImports: true,
-                noEmit: false,
-                outDir: '//'
+              minify: false, // TODO: maybe we could omit terser if `true`?
```
SukkaW, line 349:
```suggestion
              minify: false,
```

The `TODO` here should be removed. The minify shouldn't be done during bundling (otherwise the runtime code introduced by webpack can not be minimized). IMHO, if swc's minify should be used, the `TerserWebpackPlugin` should be used instead.

src/index.js
```diff
@@ -646,17 +649,17 @@ function ncc (
...
     // directory
-    if (item[""] === true) getFlatFiles(item, output, getAssetMeta, tsconfig, curPath);
+    if (item[""] === true) getFlatFiles(item, output, getAssetMeta, compilerOptions, curPath);
     // file
     else if (!curPath.endsWith("/")) {
       const meta = getAssetMeta(curPath.slice(1)) || {};
-      if(curPath.endsWith(".d.ts")) {
-        const outDir = tsconfig.compilerOptions.outDir ? pathResolve(tsconfig.compilerOptions.outDir) : pathResolve('dist');
+      if (curPath.endsWith(".d.ts")) {
+        const outDir = compilerOptions?.compilerOptions?.outDir ? pathResolve(compilerOptions.compilerOptions.outDir) : pathResolve('dist');
```
SukkaW, line 662: What is `compilerOptions?.compilerOptions` for?

src/index.js
SukkaW: The latest commit resolves the problem. The review can be marked as resolved now.

src/index.js
SukkaW: The latest commit resolves the problem. The review can be marked as resolved now.

src/index.js
SukkaW: The latest commit resolves the problem. The review can be marked as resolved now.

src/index.js
```diff
@@ -108,45 +106,36 @@ function ncc (
...
       filename: 'tsconfig.json',
     });
-    fullTsconfig = loadTsconfig(configFileAbsolutePath) || {
-      compilerOptions: {}
-    };
+    const contents = fs.readFileSync(configPath, 'utf8')
+    tsconfig = JSON5.parse(contents);
+    const baseUrl = tsconfig.compilerOptions.baseUrl;
+    resolveModules.push(pathResolve(dirname(configPath), baseUrl));
+  } catch (e) { }
```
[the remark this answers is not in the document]
kwonoj: We didn't warn anything in any case https://github.com/vercel/ncc/pull/926/files#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556L136 , I'd consider this is failsafe same as before.

kwonoj:
@styfle I've updated PR and resolved most, but can't understand test failures on the ci (https://github.com/vercel/ncc/runs/7774142054?check_suite_focus=true#step:9:303)

- This doesn't happen in local somehow
- weirdly, `coverage` seems passing (https://github.com/vercel/ncc/pull/926/files#diff-3eab028f3e7e1b3e18c3c9ab79d0a3a208aa01dea4ee5d4a9234347cc7dde2dcR4) but only normal test emits different output

any possible suggestion to dig / fix this?

src/loaders/swc-loader.js
```diff
@@ -0,0 +1,110 @@
+// Source code from https://github.com/swc-project/swc-loader/blob/master/src/index.js
+// with the following changes:
+// - swapped out `@swc/core` for `@swc/wasm`
```
styfle, line 3: Any reason to use `@swc/wasm` instead of `@swc/core`? Seems like this was the cause of the slowdown that @cb1kenobi mentioned in https://github.com/vercel/ncc/pull/926#issuecomment-1198891776

test/cli.test.js
```diff
@@ -5,22 +5,23 @@ const file = global.coverage ? "/../src/cli.js" : "/../dist/ncc/cli.js";
 
 jest.setTimeout(20000);
 
-for (const cliTest of cliTests) {
-  it(`should execute "ncc ${(cliTest.args || []).join(" ")}"`, async () => {
-    const ps = fork(join(__dirname, file), cliTest.args || [], {
+describe('cli', () => {
+  it.each(cliTests)('should execute ncc $args', async ({ args, env, expect: e, timeout }) => {
```
styfle, line 9: Anything change here or is this just formatting?

src/loaders/swc-loader.js
kwonoj: Yes, see comment (https://github.com/vercel/ncc/pull/926#issuecomment-1198902746) for the explanation.

Honestly, I do not know the exact reason. Previous work in here was using it (https://github.com/vercel/ncc/pull/777/files#diff-8c08ac17cfa691b1772e338822ad7b9cada2a78c0851239d4af55c561c943001R3), and per https://github.com/vercel/ncc/pull/777/files#diff-a992dd64a8c626fedc95d613f43602b91f6dc6339893afaf405dc05fa13c9ceaR96 I only could guess per-platform native binaries are not working well with how ncc works.

test/cli.test.js
kwonoj: Formatting to use jest's built in it.each, which provides better ergonomics / outputs.

styfle: I just wasn't sure if this could cause issues with the failing tests but seems unrelated

kwonoj: I believe this shouldn't affect test emission. Also actual failure is `unit`, `cli.test` seems passing fine.

styfle: This part seems concerning:

```
    - module.exports = eval("require")("../../../../../../../.././module");
    + module.exports = eval("require")("../../../../../../../../.././module");
```

https://github.com/vercel/ncc/runs/7774142054?check_suite_focus=true#step:9:366

That seems like its incorrect

kwonoj: Yes, more problem is I couldn't see this failure on local machine.

src/loaders/swc-loader.js
cb1kenobi: FWIW, I rebooted and pulled the latest and build times have dropped from 10 seconds down to 5.5 seconds on average. Mind you the latest ncc release builds in around 2 seconds, so there is definitely a performance regression.

test/cli.test.js
styfle: What about if you run `ncc cache clean`?

kwonoj: I think I tried, but will give few more attempt.

kwonoj: Just for the update purpose - I tried a couple like cache clean and some others, but so far no luck. I'll try to reattempt when I have some time.

styfle: @kwonoj I can reproduce locally (macOS + Node.js 16.17.0) by checking out your branch and running

```
yarn install
yarn build
yarn test
```

image

kwonoj: I tried various, but really stuck with test result inconsistences between local to CI / other machines. It is possible my machine have some weird setup cauing problems, but without figuring it out I can't reliably say PR is ready to go.

styfle: Do you have docker locally? You could try cloning this branch in a docker container to see if that reproduces.

kwonoj: I don't think I have a commercial license to install docker on the work machines. I'll think about a few, however I'm very curious why this only (yet) occurs on my machine. Wiping out whole repo & reclone even didn't make changes.

PR #927, opened 2022-05-25
chore: add `memory-fs` to the devDependencies

shogo82148:
It looks that `ncc` depends to `memory-fs` directly, but it is not in the dependencies.

https://github.com/vercel/ncc/blob/c9238cc7e7c0dfc39c0b2b9e4d556928d6dee41a/src/index.js#L6

This pull request will fix a broken build in https://github.com/vercel/ncc/pull/921

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "mailgun": "^0.5.0",
     "mariadb": "^2.0.1-beta",
     "memcached": "^2.2.2",
+    "memory-fs": "^0.5.0",
     "mkdirp": "^1.0.4",
     "mongoose": "^5.3.12",
     "mysql": "^2.16.0",


package.json
```diff
@@ -71,6 +71,7 @@
     "mailgun": "^0.5.0",
     "mariadb": "^2.0.1-beta",
     "memcached": "^2.2.2",
+    "memory-fs": "^0.5.0",
```
styfle, line 74: You'll need to run `yarn install` so that it updates the `yarn.lock` file too

shogo82148: I did.
However `yarn.lock` is not updated, because `memory-fs` already exists in `yarn.lock`.

https://github.com/vercel/ncc/blob/c9238cc7e7c0dfc39c0b2b9e4d556928d6dee41a/yarn.lock#L10208-L10214

```
$ yarn why memory-fs
yarn why v1.22.18
[1/4] 🤔  Why do we have the module "memory-fs"...?
[2/4] 🚚  Initialising dependency graph...
warning Resolution field "grpc@1.24.11" is incompatible with requested version "grpc@1.23.3"
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "memory-fs@0.5.0"
info Has been hoisted to "memory-fs"
info Reasons this module exists
   - Specified in "devDependencies"
   - Hoisted from "ts-loader#enhanced-resolve#memory-fs"
info Disk size without dependencies: "52KB"
info Disk size with unique dependencies: "352KB"
info Disk size with transitive dependencies: "584KB"
info Number of shared dependencies: 9
✨  Done in 1.78s.
```

styfle (approved):
Thanks!

PR #952, opened 2022-07-25
fix: filterAssetBase

### LINKED ISSUE #951: Bug: asset relocation doesn't work with yarn monorepos because `filterAssetBase` is set to `cwd`
Steps to reproduce:
- have a yarn workspaces monorepo
- have 2 packages in the monorepo
- package A has an asset dependency
- package B uses package A
- build package B with ncc

Expected result:
- the asset is relocated

Actual result:
- the asset is not relocated because `filterAssetBase` is set to `cwd` and the two packages are symlinked

My workaround:
- the build script changed directory to root before building

Proposed solution:
- I don't understand why `filterAssetBase` should be set to anything other than `/`. Why does it matter where the asset is located? if it's referenced original code, it should be relocated. So `filterAssetBase` should be set to `/`
- (this is related) There are many options exposed by `ncc` in javascript but not via the cli command. Have a jest/webpack style configuration javascript file, that would export all the properties in the `ncc` interface. This object would be merged to override the default configuration passed by `cli.js`.

rprovodenko:
Fixes #951

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@vercel/ncc",
   "description": "Simple CLI for compiling a Node.js module into a single file, together with all its dependencies, gcc-style.",
-  "version": "0.34.0",
+  "version": "0.34.1",
   "repository": "vercel/ncc",
   "license": "MIT",
   "main": "./dist/ncc/index.js",
--- a/src/index.js
+++ b/src/index.js
@@ -47,7 +47,7 @@ function ncc (
     assetBuilds = false,
     watch = false,
     v8cache = false,
-    filterAssetBase = process.cwd(),
+    filterAssetBase = "/",
     existingAssetNames = [],
     quiet = false,
     debugLog = false,


styfle (commented):
Thanks! 

Can you add a test to ensure this works as you expect?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@vercel/ncc",
   "description": "Simple CLI for compiling a Node.js module into a single file, together with all its dependencies, gcc-style.",
-  "version": "0.34.0",
+  "version": "0.34.1",
```
styfle, line 4: The version shouldn't be changed in PRs, it will happen during release

src/index.js
```diff
@@ -47,7 +47,7 @@ function ncc (
     assetBuilds = false,
     watch = false,
     v8cache = false,
-    filterAssetBase = process.cwd(),
+    filterAssetBase = "/",
```
styfle, line 50: Does this work on Windows?

rprovodenko:
sorry, this PR wasn't meant for merging - I was doing some experiments. I'll open another PR soon

PR #955, opened 2022-07-28
major: change filterAssetBase to root directory

rprovodenko:
What's in this PR:
- fix for https://github.com/vercel/ncc/issues/951
- `filterAssetBase` is set to current working directory's filesystem root in platform independent way
- test


rprovodenko:
@styfle hey mate, would you be able to look at this PR?

test/cli.js
```diff
@@ -1,4 +1,5 @@
 const { join } = require('path')
+const fs = require("fs");
```
styfle, line 2: Looks like this line can be removed

src/cli.js
```diff
@@ -250,7 +250,8 @@ async function runCmd (argv, stdout, stderr) {
           transpileOnly: args["--transpile-only"],
           license: args["--license"],
           quiet,
-          target: args["--target"]
+          target: args["--target"],
+          filterAssetBase: parse(process.cwd()).root
```
styfle, line 254: This seems dangerous to change the default behavior because you might include files above the current directory on accident.

Imagine you have `/Users/me/node_modules` and you run this in `/Users/me/mymonorepo/someapp`. You might accidentally include `/Users/me/node_modules`

How about exposing this as a CLI option like `--base`

rprovodenko: so this is actually how yarn monorepos work. Dependencies in sub packages are hoisted to the root packages node_modules. So it's actually how you want it to work.

styfle: When using [yarn workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/) the `filterAssetBase` should be the root of the monorepo. 

However, the code here is setting `filterAssetBase` to the root of the filesystem, which is incorrect.

PR #987, opened 2022-10-13
chore: update tsconfig-paths-webpack-plugin to 3.5.2

fiws:
This fixes a super annoying bug that can occur when `baseUrl` is set to `.` in the `tsconfig.json`. With the previous version of tsconfig-paths-webpack-plugin this would result in packages like `mongoose` not properly `require`-ing their own `package.json` but instead reading the package json of the root project.

also see https://github.com/dividab/tsconfig-paths-webpack-plugin/commit/2d838a0d7392fca3976326489cd9dfd12c3f5254


hebertcisco (approved):

styfle (approved):
Thanks!

fiws:
flaky tests?

PR #999, opened 2022-11-08
fix: node 18 by update source-map used by Terser to 0.7.4

heypiotr:
Fixes minification failing on Node 18 with:

> You must provide the URL of lib/mappings.wasm by calling SourceMapConsumer.initialize({ 'lib/mappings.wasm': ... }) before using SourceMapConsumer

https://github.com/mozilla/source-map/issues/454

There's also an open dependabot PR to bump the Terser version itself, I confirmed that also fixes the issue:
[chore(deps-dev): bump terser from 5.10.0 to 5.15.1](https://github.com/vercel/ncc/pull/988)


styfle (approved):
Thanks!

PR #1070, opened 2023-05-15
config -g are deprecated

8MiYile:


styfle (changes requested):
I don't think its deprecated.

The `--global` flag is still in the v9 docs here:

https://docs.npmjs.com/cli/v9/commands/npm-install

Also there is no mention of `--location`.

8MiYile:
@styfle 
you need use Node 18

styfle:
The version of `node` is irrelevant because the command is for `npm`

PR #1105, opened 2023-08-25
feat: add support for TypeScript 5.0's array extends in tsconfig

smorimoto:
- add support for TypeScript 5.0's array extends in tsconfig
- drop support for Node.js 14
- remove tests for Tensorflow
- update most dependencies


smorimoto:
Since some dependencies had to be updated, I updated all dependencies directly linked to ncc to the latest version and then confirmed that all tests passed.

package.json
```diff
@@ -89,25 +89,25 @@
...
     "socket.io": "^4.1.3",
     "source-map-support": "^0.5.9",
     "stripe": "^8.167.0",
     "swig": "^1.4.2",
-    "terser": "^5.6.1",
+    "terser": "^5.19.2",
     "the-answer": "^1.0.0",
     "tiny-json-http": "^7.0.2",
-    "ts-loader": "^9.3.0",
-    "tsconfig-paths": "^3.7.0",
```
smorimoto, line 102: https://github.com/dividab/tsconfig-paths/pull/245

styfle (commented):
Thanks! Can you add a test to confirm the new behavior works as expected?

smorimoto:
@styfle Done!

test/unit/tsconfig-paths-array-extends/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+	"extends": ["./tsconfig.a.json", "./tsconfig.b.json"],
```
smorimoto, line 2: > Path to base configuration file to inherit from (requires TypeScript version 2.1 or later), or array of base files, with the rightmost files having the greater priority (requires TypeScript version 5.0 or later).

That is, the priorities are:

1. `tsconfig.json`
2. `tsconfig.b.json`
3. `tsconfig.a.json`

This test verifies that the override of the `allowJs` property and the combination of A and B are working as expected.

smorimoto:
@styfle Could you check this?

.github/workflows/ci.yml
```diff
@@ -39,10 +39,10 @@ jobs:
     - name: Run Tests
       run: yarn test
     - name: Coverage
-      if: matrix.os == 'ubuntu-latest' && matrix.node == 14 # only run once
+      if: matrix.os == 'ubuntu-latest' && matrix.node == 16 # only run once
```
styfle, line 42:
```suggestion
      if: matrix.os == 'ubuntu-latest' && matrix.node == 20 # only run once
```

styfle:
Looks great, thanks! Running CI now <picture data-single-emoji=":spinner2:" title=":spinner2:">:spinner2:</picture>

smorimoto:
@styfle The latest commits should also fix issues (not directly related to this PR!) related to native builds. I squashed the commit you added and put you in as a co-author.

smorimoto:
@styfle Additionally, branch protection appears to be enabled for the already-removed Node.js 14. Don't forget to update that as well!

smorimoto:
@styfle Could you kick the CI?

styfle:
Running CI again…

smorimoto:
Oh... Windows is always a nightmare...

smorimoto:
@styfle One last poke...

smorimoto:
I think you should definitely update this configuration...

image

smorimoto:
@styfle I temporarily removed tfjs from the test as it was too unstable to break on other platforms when I fixed it for either platform. We can come back here again when it's correct in upstream.

smorimoto:
@styfle Finally, all the tests are expected to pass, and that can be confirmed [here](https://github.com/smorimoto/ncc/actions/runs/6078870706). Now all we have to do is make sure the CI is OK here and go ahead.

.github/workflows/ci.yml
```diff
@@ -20,29 +20,34 @@ jobs:
...
+        python-version: '3.11'
     - name: Use Node.js ${{ matrix.node }}
       uses: actions/setup-node@v3
       with:
         cache: yarn
         node-version: ${{ matrix.node }}
         check-latest: true
     - name: Install Dependencies
-      run: yarn install
+      run: yarn global add node-gyp && yarn install --frozen-lockfile
```
styfle, line 39: Why does node-gyp need to be a global? 

Can we add it as a `devDependency` instead?

styfle:
Please update the PR description with example usage and link any existing issue, thanks!

.github/workflows/ci.yml
smorimoto: It's already listed there, but somehow it seems that Windows sometimes needs it in a global environment. I haven't investigated whether this is a problem with the package manager resolver or some kind of implementation, but generally there shouldn't be a major problem

styfle (approved):
Thanks!

smorimoto:
@styfle Whoa! Thank you for taking care of the rest of the work!

PR #1122, opened 2023-10-12
fix: sourcemap sources removes webpack path

### LINKED ISSUE #1011: ncc producing faulty source maps?
I'm having to manually transform the output `.js.map` file after building because otherwise it does not work with c8. The generated `.map` file has invalid file paths that cause c8 to not work correctly. The faulty paths begin with something like `../webpack://` (clearly an invalid filesystem path).

Now, I was ultimately able to get my coverage reports working correctly after a couple days of constant debugging and headaches. The solution I came up with is to manually transform the `.map` file post-build/post-compile in order to fix the faulty paths. The faulty paths appear in the `sources` key of the `.js.map` file. You can view my temporary fix here: https://github.com/tmillr/breaking-change/blob/c432a9474cf9f5ee4a47f9945583fbb5a9869683/scripts/fix-sourcemap-sources.js#L1-L13

What I learned (I think), is that c8 uses these generated paths in the `sources` key of the `.map` file in order to retrieve the actual (pre-compile) source content from the filesystem. So, when these filesystem paths are incorrect/malformed because they begin with `../webpack://`, c8 will produce inaccurate/faulty coverage reports (in my case, in my tests, it was being reported (via c8) that the compiled file was never even loaded/run, even though it was, and to make matters even worse, c8 was not making it obvious that there was any kind of error).

### LINKED ISSUE #1121: Source maps are generated incorrectly and cannot be debugged in vscode.
I want to debug code with breakpoints in vscode. vscode doesn't recognize breakpoints after generating a .map file with `ncc build ./src/index.ts -o dist --source-map --no-source-map-register` . Looking at the .map file, the `sources` option in there is 
```json
"sources":["../webpack://typescript/webpack/bootstrap","../webpack://typescript/./index.ts"]
```
manually change to 
```json
"sources":["../webpack/bootstrap","../index.ts"],
``` 
and then vscode can debug at breakpoints. 

Looking at the commit history, the sources entries for .map files generated before [c7b461c](https://github.com/vercel/ncc/blob/c7b461ceb4636232073ddecfc8be88bc92f97666/test/unit/minify/output.js.map) are correct

axuj:
- Fixes #1011 
- Fixes #1121 

The source is now in the format `webpack://[namespace]/[resourcePath]`


src/index.js
```diff
@@ -475,22 +475,11 @@ function ncc (
...
-        if (source.startsWith('external '))
-          source = 'node:' + source.slice(9);
-        if (source.startsWith('./'))
-          source = source.slice(2);
-        if (source.startsWith('(webpack)'))
-          source = 'webpack' + source.slice(9);
-        if (source.startsWith('webpack/'))
-          return '/webpack/' + source.slice(8);
-        return sourceMapBasePrefix + source;
+        // webpack://[namespace]/[resourcePath]
```
styfle, line 478: Looks like this no longer accounts for the `webpack:///webpack:///` case.

axuj: Yes, look at [output.devtoolModuleFilenameTemplate](https://webpack.js.org/configuration/output/#outputdevtoolmodulefilenametemplate) , I think using webpack config is a better choice

styfle: Thanks! Is there a way to add a test to confirm this fixes the issue?

axuj: Sure! I've added tests

styfle (approved):
Thanks! 🎉

PR #1155, opened 2024-01-29
feat: add `webpack` arg for customizing webpack config

iamhollow:
This PR enables the base ncc webpack config to be customized or extended, similar to [NextJS](https://nextjs.org/docs/app/api-reference/next-config-js/webpack):

https://nextjs.org/docs/app/api-reference/next-config-js/webpack

The use cases of this feature are scenarios such as adding additional webpack rules or plugins, or configuring a path resolve alias.

The README has been updated for the programmatic invoke from Node.js to include the new `webpack` arg.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -128,7 +128,7 @@ require('@vercel/ncc')('/path/to/input', {
   v8cache: false, // default
   quiet: false, // default
   debugLog: false, // default
-  webpack: (config) => config // default
+  customizeConfig: (config) => config // default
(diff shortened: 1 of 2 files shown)


styfle (changes requested):
Thanks for the PR!

We don't want to expose anything specific to webpack because that will make it much harder to swap webpack to another bundler (such as turbopack) in the future.

iamhollow:
@styfle I renamed the exposed variable to a more generic wrapper called "customizeConfig"

PR #1192, opened 2024-05-24
fix: Update build process dependencies

ilnicki:
Current version of ncc refuses to build my project with not really helpfull error like:
`node_modules/@vercel/ncc/dist/ncc/index.js.cache.js:73627
                      resource: j.length > 1 ? j.map((E) => E.path) : j[0].path
TypeError: Cannot read properties of undefined (reading 'path')`
I've formatted index.js.cache.js for message to be more meaningfull.

Updating building dependencies fixes it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "@google-cloud/firestore": "^6.7.0",
     "@sentry/node": "^7.66.0",
     "@slack/web-api": "^6.9.0",
-    "@vercel/webpack-asset-relocator-loader": "1.7.3",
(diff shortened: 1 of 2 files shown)


styfle (commented):
Thanks for the PR!

Can you add a test to verify this fixes the bug?

JulienKode:
Thanks for taking a look at this @ilnicki

macrozone:
i have the same error, can we get this merged? not sure if it fixes my error though

PR #1194, opened 2024-06-04
fix: upgrade to `terser@5.31.0` to fix minfy with negated `in` statements

### LINKED ISSUE #1193: Outdated Terser causes a syntax issue when minify-ing code
We are trying to swap out `node-fetch` for `undici`, but unfortunately we are running into ncc compiling issues. Specifically, there is a patched issue with Terser removing parentheses for `!(#test in this)`, causing the syntax issue.

- The issue was fixed in Terser starting from `terser@5.30.3`, see https://github.com/terser/terser/issues/1516.
- The code that triggered this issue can be found in `undici`'s code itself: https://github.com/nodejs/undici/blob/2c9bc0b74f914a120a0497efad791b471767dc55/lib/web/fetch/util.js#L882C7-L882C75

Is it possible to release a patch version with a more up-to-date Terser, to include this fix?

byCedric:
Fixes #1193

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "source-map-support": "^0.5.9",
     "stripe": "^8.167.0",
     "swig": "^1.4.2",
-    "terser": "^5.19.3",
+    "terser": "^5.31.0",
     "the-answer": "^1.0.0",
     "tiny-json-http": "^7.5.1",
     "ts-loader": "^9.4.4",
--- a/yarn.lock
+++ b/yarn.lock
@@ -14965,10 +14965,10 @@ terser-webpack-plugin@^5.1.3:
     source-map "^0.6.1"
     terser "^5.7.0"
 
-terser@^5.19.3:
-  version "5.19.4"
-  resolved "https://registry.yarnpkg.com/terser/-/terser-5.19.4.tgz#941426fa482bf9b40a0308ab2b3cd0cf7c775ebd"
-  integrity sha512-6p1DjHeuluwxDXcuT9VR8p64klWJKo1ILiy19s6C9+0Bh2+NWTX6nD9EPppiER4ICkHDVB1RkVpin/YW2nQn/g==
+terser@^5.31.0:
+  version "5.31.0"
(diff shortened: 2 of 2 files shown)


styfle (commented):
Thanks for the PR! Looks like CI is failing however.

https://github.com/vercel/ncc/actions/runs/9369710285/job/25854934076?pr=1194#step:9:335

byCedric:
These are interesting failures, I'm not sure why Node 20 on Ubuntu causes that error, while Node 20 on Windows seems fine.

The MacOS errors are just `node-gyp` (install) related issues, likely due to the GHA runner missing some low-level dependency.

byCedric:
Also interesting here is that the last run on workflow (4 months ago) did not succeed for Node 20 on Ubuntu ([this run](https://github.com/vercel/ncc/actions/runs/7873890228)). Unfortunately, the logs have been wiped already, but that might be a pointer to this failure too.

I'm also guessing the `macos-latest` runner has been changed within 4 months, causing the CI failures.

styfle:
I see what you mean. I'll run CI against main to see how it compares:

https://github.com/vercel/ncc/pull/1195

byCedric:
Glad we you could repro this on `main` (as in, not caused by this PR) 😄

Anything I can do to help?

styfle:
I think we need to fix main before merging this to confirm it works as expected

theoludwig:
> I think we need to fix main before merging this to confirm it works as expected

What was the problem with CI (the logs disappeared)? Should probably run CI again, with all the depdencies updated to their latest version, to see the status of the project.

styfle:
@theoludwig Feel free to create a new PR to fix it

theoludwig:
> @theoludwig Feel free to create a new PR to fix it

Alright. :smile: https://github.com/vercel/ncc/pull/1213

Tests are working locally, but about CI, I don't know, need an approval to run it. Hopefully, it fixes the issue, and allow to update terser and webpack to latest version to get latest fixes like this one with negated `in` statements.

PR #1213, opened 2024-09-10
fix(deps): update webpack to v5.94.0, terser to v5.33.0

### LINKED ISSUE #1193: Outdated Terser causes a syntax issue when minify-ing code

theoludwig:
Attempt to fix CI, and update webpack and terser to their latest version.

- Fixes #1193
- Closes #1194
- Closes #1177
- Closes #1204
- Closes #1195


styfle (commented):
Looks like macos is failing to install dependencies. 

We might need to update `canvas` as well.

theoludwig:
> Looks like macos is failing to install dependencies.
> 
> We might need to update `canvas` as well.

Thank you for running CI, and helping for fixing this issue. :smile:

I've updated canvas to https://github.com/Automattic/node-canvas/releases/tag/v3.0.0-rc2, hopefully it fixes issues with macOS, but for "Node 20 and ubuntu-latest", it seems like there is another issue.

styfle:
Looks like updating canvas didn't help. Perhaps we need to install `pangocairo` maybe others:

https://github.com/Automattic/node-canvas/issues/1050#issuecomment-351080589

styfle:
Oh yeah, you can see their CI has it here:

https://github.com/hargasinski/node-canvas/blob/e7abe64833d13ec96449c827b1e14befbdf3105d/.github/workflows/ci.yaml#L68-L70

package.json
```diff
@@ -106,7 +106,8 @@
     "vue": "^2.5.17",
     "vue-server-renderer": "^2.5.17",
     "web-vitals": "^0.2.4",
-    "webpack": "5.76.0",
+    "webpack": "5.94.0",
     "when": "^3.7.8"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
theoludwig, line 112: Ah, `corepack` added it automatically. :laughing: 
Is it fine, or should I remove it?

styfle: Let's keep it. Eventually we'll want to move off yarn 1 but its good to document which version was used.

test/integration.test.js
```diff
@@ -110,13 +113,7 @@ it(`should execute "ncc build web-vitals" with target config`, async () => {
   await new Promise(resolve => setTimeout(resolve, 5000));
 });
 
-afterAll(() => {
-  if (coverage)
-    process.exit(0);
-});
```
theoludwig, line 116: Why is this necessary? I think it's causing issues in CI, I removed it for now.

test/integration.test.js
```diff
@@ -44,6 +44,9 @@ for (const integrationTest of fs.readdirSync(__dirname + "/integration")) {
   // ignore a few tests known to fail on windows
   if (process.platform === 'win32' && skipOnWindows.includes(integrationTest)) continue;
 
+  // ignore leveldown.js test on macOS, https://github.com/Level/leveldown/issues/801
+  if (process.platform === 'darwin' && integrationTest === 'leveldown.js') continue;
```
theoludwig, line 48: It is the only test failing for macOS, might make sense to disable it? Might be related to https://github.com/Level/leveldown/issues/801.

test/integration.test.js
styfle: I don't recall why this was here. It was probably to fix a bug where the process was hanging. We'll see if it hangs once removed.

test/integration.test.js
styfle: Sounds good to me 👍

test/integration.test.js
styfle: Looks like all passed except the test coverage 

https://github.com/vercel/ncc/actions/runs/10862816229/job/30202179532?pr=1213

Let's add this back

theoludwig: Alright, I added it back. Hopefully, all CI issues are resolved, but I feel like it will still not work for "Node 20 and ubuntu-latest", with test coverage.

test/integration.test.js
theoludwig: (I've refactored the skip, to be the same way as for the tests skipped on Windows)

styfle:
Are you able to run coverage locally?

The error looks like:

```
  ● should generate correct output for browser-mainfield
    TypeError: Cannot read properties of undefined (reading 'then')
      at node_modules/webpack/lib/FileSystemInfo.js:1908:19
      at processQueue (node_modules/webpack/lib/util/processAsyncTree.js:56:4)
```

theoludwig:
When I first started this PR, I was able to run `yarn test` locally without issues, then I focused on the CI problems, and haven't ran tests locally since.
I never tried `yarn test-coverage` locally, my bad, I should have ran it, it seems like it is also part of the main issue since the beginning.

> Perhaps you could start with a new PR that doesn't upgrade webpack and only gets CI passing?

Yeah, that's a good idea, to start simple and "small" as possible. Priority is to first fix CI, then we'll see.
However, I tried `rm -rf node_modules .cache tmp`, `git checkout main`, `yarn`, `yarn test`, and the same issue seem to appear:
```js
      at node_modules/webpack/lib/FileSystemInfo.js:1676:22
      at processQueue (node_modules/webpack/lib/util/processAsyncTree.js:50:4)
/home/theoludwig/Documents/open-source/ncc/node_modules/webpack/lib/FileSystemInfo.js:1677
							lexer.init.then(() => {
							           ^

TypeError: Cannot read properties of undefined (reading 'then')
    at /home/theoludwig/Documents/open-source/ncc/node_modules/webpack/lib/FileSystemInfo.js:1677:19
    at processQueue (/home/theoludwig/Documents/open-source/ncc/node_modules/webpack/lib/util/processAsyncTree.js:50:4)
    at processTicksAndRejections (node:internal/process/task_queues:77:11)
```

---

I will check later, when I have time, what could be the root cause of the issue.

styfle:
Yes, ideally you could run `yarn test-coverage` locally and not have to wait on me to approve CI. That should help you iterate faster. Thanks for looking into it!

theoludwig:
> Yes, ideally you could run `yarn test-coverage` locally and not have to wait on me to approve CI. That should help you iterate faster. Thanks for looking into it!

Yep! (was just blindlessly thinking it was only related to CI config problems :sweat_smile:)

After debugging for a while, I found that the error might be related to a race condition happening when `process.exit(0);` is called in coverage mode in `integration.test.js` but `watcher.test.js` tests are still running.

I first tried to run each test file separately, like: `yarn test-coverage test/integration.test.js`, `yarn test-coverage test/watcher.test.js`, etc. And everything was working fine locally.

So I tried with [jest `--runInBand` CLI option](https://jestjs.io/docs/cli#--runinband), and now `yarn test-coverage` works locally. The `leveldown.js` test on macOS still need to be skipped, and OS dependencies still need to be installed.

Just to be 100% sure, that the CI problem is resolved, I enabled GitHub Actions on my fork, and ran it here: https://github.com/theoludwig/ncc/pull/1

And good news, it is working:
image

The PR is here: https://github.com/vercel/ncc/pull/1216
It's focusing only on fixing the CI and tests, when it is merged, I will update this one, to update webpack and terser (most importantly, terser, if the webpack update doesn't work).

styfle:
@theoludwig Nice find! 🕵️ 

The other PR was merged so feel free to update this one, thanks! 🎉

styfle (approved):
Great work, thanks! 🎉

PR #1215, opened 2024-09-17
Update readme.md

gregg-cbs:
Adding the need to install ts-loader otherwise typescript builds fail and throw an error.


styfle (changes requested):
You don't need to install it, it comes with ncc

gregg-cbs:
i had to install it separately, im not sure what else to say. I got an error otherwise.

PR #1236, opened 2024-11-19
fix(cjs-build): enable evaluating import.meta in cjs build

### LINKED ISSUE #1019: Cannot use 'import.meta' outside a module on 0.36.0 (not 0.34.0)
I just updated from version 0.34.0 to 0.36.0. In version 0.34.0 it works perfectly.

```
ncc: Version 0.36.0
ncc: Compiling file index.js into CJS
...
esm_dirname = (0,external_url_.fileURLToPath)(import.meta.url);
                                                       ^^^^

SyntaxError: Cannot use 'import.meta' outside a module
...
```

nicotsx:
## Purpose

A while ago, a [PR](https://github.com/vercel/ncc/pull/897) was merged to disable `import.meta` evaluation altogether to support new versions of Webpack. This sometimes causes issues in CJS projects.

In the comments, there were some [discussions](https://github.com/vercel/ncc/pull/897#discussion_r836916315) about whether we should have a conditional and keep the previous behavior for CJS builds. This PR addresses exactly this by translating `import.meta.url` to a CJS compatible statement which results to the same output. 

I have created a [comparative build](https://github.com/meienberger/ncc/pull/1/files) between two projects using the same code and a dependency (@sentry/node) that uses `import.meta.url`. You can see the difference in the output `index.js`

Before the change:

```javascript
const importMetaUrl = typeof import.meta.url !== 'undefined' ? import.meta.url : undefined;
```

After:

```javascript
const importMetaUrl = typeof require("url").pathToFileURL(__filename).href !== 'undefined' ? require("url").pathToFileURL(__filename).href : undefined;
```

Closes #1019


styfle (commented):
Thanks for the PR! 

Please add a test to confirm this is working as expected.

styfle (approved):
Great work, thanks!

PR #1248, opened 2025-02-11
node target fixes #1247

### LINKED ISSUE #1247: would you be open to support node[[X].Y] as target to make GitHub Action build artifacts smaller?
in the light of 1246 I discovered default target is es2025 and it only es* targets are expected. underlying webpack does support node* as a target, but ncc restricts it. but using es version as a target doesn't make sense for nodejs versions, because es support is fragmented.

I'd like to relax es* only target restriction to include node* as well, to make GitHub Action build artifacts smaller by not transpiling everything to es2015 but rather only to what's missing in node20 for example

iamstarkov:


iamstarkov:
> Note that webpack runtime code is not the same as the user code you write, you should transpile those code with transpilers like Babel if you want to target specific environments, e.g, you have arrow functions in source code and want to run the bundled code in ES5 environments. Webpack won’t transpile them automatically with a targetconfigured.
> — https://webpack.js.org/configuration/target/#target

maybe this the reason why why nodeX.Y isnt doing transpiling

PR #1254, opened 2025-02-16
fix: make sourcemap base prefix relative to the build file instead of default "../" (fixes #1253)

### LINKED ISSUE #1253: sourcemap filesystem mapping is broken when output is more than one folder deep
in my case I have a GitHub action which can consists of smaller sub actions which can be used by themselves. Think of [`actions/cache`](https://github.com/actions/cache) with its [`actions/cache/save`](https://github.com/actions/cache/tree/main/save) and [`actions/cache/restore`](https://github.com/actions/cache/tree/main/restore). cache action builds into single [dist folder but have multiple sub folders](https://github.com/actions/cache/tree/main/dist) for every sub action. I do the same.

Sourcemaps work, but location mapping is one off in terms of traversing up the filesystem hierarchy. I looked into the code and it seems despite ncc programmatically supporting `sourceMapBasePrefix`, it is not computed at all, let alone relatively to the `--output`. when I add on extra, sourcemap locations start to match to the original files and I can click on the stack trace and get to an error directly from my vs code

```
~/projects/nordnet/gitops (fix/multi-document-yaml) $ npm run build:pr

> gitops@0.0.0 build:pr
> ncc build pr/action.js -o dist/pr/ --source-map

ncc: Version 0.38.3
ncc: Compiling file index.js into ESM
   0kB  dist/pr/package.json
  40kB  dist/pr/sourcemap-register.cjs
1467kB  dist/pr/index.js
1715kB  dist/pr/index.js.map
1715kB  dist/pr/index.js.map
3222kB  [1237ms] - ncc 0.38.3
```

```
// before fix
~/projects/nordnet/gitops (fix/multi-document-yaml) $ node --env-file=.env.test dist/pr/index.js

file:///Users/vlasta/projects/nordnet/gitops/dist/src/utils/enhanced-core.js:7
    .foo()
^
TypeError: str.foo is not a function
    at getListFromString (file:///Users/vlasta/projects/nordnet/gitops/dist/src/utils/enhanced-core.js:7:1)
    at Object.getListInput (file:///Users/vlasta/projects/nordnet/gitops/dist/src/utils/enhanced-core.js:26:1)
    at run (file:///Users/vlasta/projects/nordnet/gitops/dist/pr/action.js:35:1)


// after fix
~/projects/nordnet/gitops (fix/multi-document-yaml) $ node --env-file=.env.test dist/pr/index.js

file:///Users/vlasta/projects/nordnet/gitops/src/utils/enhanced-core.js:7
    .foo()
^
TypeError: str.foo is not a function
    at getListFromString (file:///Users/vlasta/projects/nordnet/gitops/src/utils/enhanced-core.js:7:1)
    at Object.getListInput (file:///Users/vlasta/projects/nordnet/gitops/src/utils/enhanced-core.js:26:1)
    at run (file:///Users/vlasta/projects/nordnet/gitops/pr/action.js:35:1)
```

iamstarkov:


styfle (commented):
Thanks, can you add a test to confirm this works?

iamstarkov:
> Thanks, can you add a test to confirm this works?

Yes, I'll give it a try

iamstarkov:
@styfle added relevant tests

some unrelated tests fail for me locally both in my branch and in up to date main branch.

Here is the tests and corresponding errors:
● should execute "ncc run aws-sdk.js": `Error: Cannot find module '@aws-sdk/client-s3'`
● should execute "ncc run binary-require.js": `Error: Cannot find module './hello.node'`
● should execute "ncc run canvas.js": `Error: Cannot find module '../build/Release/canvas.node'`

I assume its because `npm run build-test-binary` fails due to missing `distutils`, which had been [fixed in node-gyp 10, while ncc uses 9](https://github.com/nodejs/node-gyp/issues/2869)

iamstarkov:
how does it look?

src/index.js
```diff
@@ -295,7 +295,11 @@ function ncc (
       libraryTarget: esm ? 'module' : 'commonjs2',
       strictModuleExceptionHandling: true,
       module: esm,
-      devtoolModuleFilenameTemplate: sourceMapBasePrefix + '[resource-path]'
+      // if prefix is `../`, then final path results in `.././`
+      // if prefix is `../../`, then final path results results in `../.././`
+      // which ideally would have been normalized by webpack, but alas it isnt,
+      // so we need to adjust sourceMapBasePrefix
+      devtoolModuleFilenameTemplate: sourceMapBasePrefix.slice(0, -1) + '[resource-path]'
```
styfle, line 302:
```suggestion
      devtoolModuleFilenameTemplate: sourceMapBasePrefix.endsWith('/') ? sourceMapBasePrefix.slice(0, -1) : sourceMapBasePrefix + '[resource-path]'
```

This should be conditional right? What if you accidentally remove the filename?

styfle: Actually lets go with

```js
sourceMapBasePrefix.replace(/\/$/, '') + '[resource-path]'
```

iamstarkov:
```suggestion
      devtoolModuleFilenameTemplate: sourceMapBasePrefix.replace(/\/$/, '') + '[resource-path]'
```

iamstarkov: @styfle applied the change, pls re-review

styfle (commented):
Thanks for the PR!

Looks like tests are failing CI though.

Were you able to get tests passing locally?

iamstarkov:
> Thanks for the PR!
> 
> Looks like tests are failing CI though.
> 
> Were you able to get tests passing locally?

last time around all but few unrelated tests were passing, see more in the prev comment https://github.com/vercel/ncc/pull/1254#issuecomment-2838920504

iamstarkov:
I believe the test dont work no more, because suggested change broke them =)

PR #1258, opened 2025-03-13
perf: use swc for minify

SukkaW:
Unlike #777 and #926 where swc is used to replace `ts-loader`, in this PR swc is used to replace `terser` in the modification phase.


test/unit/minify-sourcemap-register/output.js
```diff
@@ -1,2 +1 @@
 require('./sourcemap-register.js');(()=>{if(typeof __nccwpck_require__!=="undefined")__nccwpck_require__.ab=__dirname+"/";var e={};(()=>{var _=e;const o="qux";console.log?.("hello");_.foobar=o})();module.exports=e})();
-//# sourceMappingURL=index.js.map
```
styfle, line 2: This looks like a bug since the source map comment is missing

styfle:
> Unlike https://github.com/vercel/ncc/pull/777 and https://github.com/vercel/ncc/pull/926 where swc is used to replace ts-loader, in this PR swc is used to replace terser in the modification phase

I think this is only worth doing if we also plan to replace swc for the `ts-loader` phase too.

Otherwise we include a massive dependency that we're not taking advantage of.

- https://packagephobia.com/result?p=@swc/wasm@1.11.9
- https://packagephobia.com/result?p=terser@5.39.0

Also did you compare performance before/after?

SukkaW:
> Otherwise we include a massive dependency that we're not taking advantage of.

We can always do this gradually since #926 has been blocked and simply abandoned since (no one is investigating it anymore).

> Also did you compare performance before/after?

I use the integration test case (`yarn build && time yarn test test/integration.test.js`) to compare the performance.

Before: `yarn test test/integration.test.js  167.13s user 15.04s system 246% cpu 1:14.01 total`
`@swc/wasm`: `yarn test test/integration.test.js  150.22s user 15.37s system 234% cpu 1:01.04 total`
`@swc/core`: `yarn test test/integration.test.js  149.66s user 15.27s system 249% cpu 1:00.60 total`

However integration test cases use many packages anyway, thus I assume the performance improvement is negligible on small projects with fewer dependencies.

test/unit/minify-sourcemap-register/output.js
SukkaW: Seems to be a limitation of swc as it is impossible to pass an object to the `minify.sourceMap` option.

https://github.com/vercel/ncc/blob/36a7568175b5938353f3171b17bda68ebe025d06/src/index.js#L485-L489

cc @kdy1

kdy1: Should SWC write `//# sourceMappingURL=index.js.map`?

https://github.com/swc-project/swc/blob/a53c60db18bd90539f3449348cbb166f32f46223/crates/swc_compiler_base/src/lib.rs#L259-L271

SWC has relevant code for sourceMap = `inline` but not for `true`. Does terser emit one?

kdy1: https://terser.org/docs/api-reference/#source-map-options so there was an option. I'll implement this option today or tomorrow and publish a new version of `@swc/core`

kdy1: https://github.com/swc-project/swc/pull/10346 implements the option, and I'll publish a new version later today, after merging some PRs.

Sorry for the late response! The mention email was buried due to lots of other emails. (Mostly PR review requests)

SukkaW: Sure. Lemme see if the tests passed here after bumping the swc to the latest version later today.

kdy1: A new version is published! Can you try updating `@swc/core`?

SukkaW: @kdy1 Just bump the swc to the latest version, and it appears to be working, thanks for fixing this! cc @styfle 

See also CI test runs: https://github.com/SukkaW/ncc/pull/2

test/unit/minify-sourcemap-register/output.js.map
```diff
@@ -1 +1 @@
-{"version":3,"file":"index.js","names":["__webpack_require__","ab","__dirname","foobar","console","log","exports"],"sources":["../webpack/runtime/compat","../test/unit/minify-sourcemap-register/input.js"],"sourcesContent":["\nif (typeof __webpack_require__ !== 'undefined') __webpack_require__.ab = __dirname + \"/\";","const foobar = 'qux'\nconsole.log?.(\"hello\");\nexports.foobar = foobar\n"],"mappings":"MACA,UAAAA,sBAAA,YAAAA,oBAAAC,GAAAC,UAAA,I,uBCDA,MAAAC,EAAA,MACAC,QAAAC,MAAA,SACAC,EAAAH,Q","ignoreList":[]}
\ No newline at end of file
+{"version":3,"file":"index.js","sources":["../webpack/runtime/compat",".././test/unit/minify-sourcemap-register/input.js"],"sourceRoot":"","sourcesContent":["\nif (typeof __webpack_require__ !== 'undefined') __webpack_require__.ab = __dirname + \"/\";","const foobar = 'qux'\nconsole.log?.(\"hello\");\nexports.foobar = foobar\n"],"names":[],"mappings":"MACA,GAAA,OAAA,sBAAA,YAAA,oBAAA,EAAA,CAAA,UAAA,2BCDA,MAAA,EAAA,MACA,QAAA,GAAA,GAAA,SACA,EAAA,MAAA,CAAA"}
```
styfle, line 1: @kdy1 Is it expected that `names` changed to an empty array?


Before:


```js
["__webpack_require__","ab","__dirname","foobar","console","log","exports"]
```

After:
```js
[]
```

That looks like a bug.

kdy1: I think it's https://github.com/swc-project/swc/issues/6726

kdy1: It's now fixed. Can you try updating `@swc/cli`?

PR #1314, opened 2026-04-02
fix(security): Predictable global cache directory in /tmp enables symlink/hijack risks

tuanaiseo:
## Problem

The cache directory is a fixed, shared path (`/tmp/ncc-cache`). On multi-user systems this can be pre-created or manipulated by another user (symlink/hardlink attacks), potentially causing cache poisoning, unintended file writes, or data leakage between builds/users depending on how cache files are later written.

**Severity**: `medium`
**File**: `src/utils/ncc-cache-dir.js`

## Solution

Use a per-user/per-project cache directory with secure creation semantics (e.g., `fs.mkdtemp` under user-owned base dir), enforce permissions (`0700`), and verify path is not a symlink before writing.

## Changes

- `src/utils/ncc-cache-dir.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced


src/utils/ncc-cache-dir.js
```diff
@@ -1 +1,8 @@
-module.exports = require("os").tmpdir() + "/ncc-cache";
\ No newline at end of file
+const crypto = require("crypto");
+const os = require("os");
+const path = require("path");
+
+const cacheBase = process.env.XDG_CACHE_HOME || path.join(os.homedir(), ".cache");
+const projectKey = crypto.createHash("sha256").update(process.cwd()).digest("hex").slice(0, 12);
```
styfle:
```suggestion
const projectKey = crypto.hash("sha1", process.cwd());
```

No reason to truncate. Just use a smaller hashing function if you don't need the entropy.

styfle (approved):
Thanks!

Timer (approved):

PR #1315, opened 2026-04-23
fix: Reorder extension resolution to prioritise TypeScript over JSON

PierreVanobbergen:
When a directory contains both `test.json` and `test.ts`, an extensionless import like:

```ts
import { myFunction } from './test';
```

resolves to `test.json` instead of `test.ts`, causing all exported methods from the TypeScript file to be unresolvable during the build.

Webpack resolves extensions in the order they appear in the `resolve.extensions` array.

Since `.json` appears before `.ts` and `.tsx`, webpack picks the JSON file first when both exist.

### Reproduce

(The test project need a `tsconfig.json` with `compilerOptions.resolveJsonModule: true`)

1. Create a directory with two files sharing the same base name: 

   **`src/test.json`**
   ```json
   { "key": "value" }
   ```

   **`src/test.ts`**
   ```ts
   export function greet(): string {
     return "hello";
   }
   ```

2. Import without an extension from another TypeScript file:

   **`src/index.ts`**
   ```ts
   import { greet } from './test';
   console.log(greet());
   ```

3. Run `ncc build src/index.ts` the build fails because `./test` resolves to `test.json`, which has no `greet` export.


styfle (commented):
Thanks! Can you add a test to confirm this works as you expect?

PierreVanobbergen:
Added a new unit test to check the resolution order modification.
I tested before and after the fix, and it behaves correctly!

styfle (approved):
Great, thanks! 🎉

styfle:
We have a requirement for git commit signing to merge so you'll need to sign your commits.

https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

I tried to merge the latest commit but its not just latest that needs signing - all commits in the PR must be signed to merge. So you could rebase into one signed commit if you want.

PierreVanobbergen:
I squashed all the commits to a signed one 👍

PR #1319, opened 2026-05-14
fix: preserve TypeScript imports for ESM builds

Jaksenc:
## Problem

Issue #1311 reports `Package path . is not exported` when ncc bundles a TypeScript entry that imports an ESM-only package such as `@actions/core@3.0.0`.

`@actions/core@3.0.0` exposes its package root through the `import` condition, but not through `require`. In a TypeScript project with `type: "module"` and `compilerOptions.module` set to `CommonJS`, ncc selects ESM output, but ts-loader lowers the entry import to `require()`. Webpack then resolves the dependency through CommonJS conditions and misses the package's `import` export.

## Change

- preserve TypeScript ES imports when ncc has selected ESM output but `tsconfig.json` asks TypeScript to emit CommonJS
- leave user TypeScript module settings alone for non-ESM builds and other module modes
- add a regression test for an ESM package that exposes only an `import` condition

The fixture sets `esm: true` explicitly so the test focuses on ncc's ESM output path while keeping the TypeScript `module: "commonjs"` setting that triggers the regression.

This patch keeps TypeScript imports intact only for that ESM-output/CommonJS-emit mismatch, so webpack can resolve import-only packages through their `import` condition without broadening the behavior for CJS ncc builds or other TypeScript module modes.

## Research / Repro

- Confirmed `@actions/core@3.0.0` exposes package root through `import` only via `npm view @actions/core@3.0.0 version type exports --json`.
- Confirmed the fixture keeps `compilerOptions.module: "commonjs"` while ncc receives `esm: true`, which is the mismatch covered here.
- Ran the new repro against `origin/main` in a detached worktree: it fails with `Package path . is not exported` for an import-only package.
- Ran the same repro against this branch: it passes and the bundled code contains the import-only package value.
- Checked open PR overlap for #1311; only this PR currently targets it.

## Validation

- `pnpm install --frozen-lockfile`
- `node scripts/build.js --no-cache`
- `./node_modules/.bin/tsc --noEmit -p test/unit/ts-esm-import-condition/tsconfig.json`
- `pnpm test -- test/unit.test.js -t ts-esm-import-condition`
- `pnpm test -- test/unit.test.js`
- `pnpm build-test-binary`
- `pnpm test -- test/integration.test.js`

## Risk

The behavior change is limited to ncc ESM builds where TypeScript would otherwise emit CommonJS. CJS ncc builds and other TypeScript module modes are left unchanged.

The new fixture includes a small package under `test/unit/ts-esm-import-condition/node_modules` to model an import-only package export map; it is static test data, not a new dependency.

## Out of scope

- changing TypeScript module settings for non-ESM ncc builds
- adding a general compatibility layer for direct `tsc + node` execution of this fixture shape
- changing dependency resolution behavior outside this ESM-output/CommonJS-emit mismatch

## Issue linkage

Addresses https://github.com/vercel/ncc/issues/1311.


src/index.js
```diff
@@ -635,6 +643,14 @@ function ncc (
   }
 }
 
+function isCommonJsModule(module) {
+  if (typeof module === 'string') {
+    return ['commonjs', 'amd', 'umd', 'system'].includes(module.toLowerCase());
+  }
+
+  return module >= 1 && module <= 4;
```
styfle: Where are these integers defined? Lets link to the docs so its clear.

test/unit.test.js
```diff
@@ -79,3 +80,60 @@ for (const unitTest of fs.readdirSync(`${__dirname}/unit`)) {
     )
   });
 }
+
+async function expectEsmOnlyTypeScriptBuild(compilerOptions) {
+  const testDir = fs.mkdtempSync(path.join(__dirname, "tmp-esm-ts-cjs-"));
```
styfle: Why is this a temp directory? Looks like you're building a test fixture so lets just create a new test fixture and commit it since this doesn't need any dynamic behavior

src/index.js
Jaksenc: Ah yeah, that’s fair. I added a link to the TypeScript enum here and renamed the helper so it’s clearer what those numbers are doing.

test/unit.test.js
Jaksenc: Youre right that makes more sense, just changed it. Thanks!!

test/unit/ts-esm-import-condition/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
```
styfle, line 3: This is super strange behavior since you have module=commonjs in tsconfig.json here but type=module in package.json

Does this code even compile and run using tsc + node?

I dont understand why ncc should support this 🤔

Jaksenc: No you’re so right, the fixture makes the case look a bit weird.

I was trying to test ncc’s ESM output path instaed of saying that this setup should be run directly with tsc + Node. Lemme clean it up by moving the import-only package into node_modules and making esm: true explicit in the fixture options so the test is clearer.

src/index.js
```diff
@@ -635,6 +643,10 @@ function ncc (
   }
 }
 
+function emitsCommonJsModule(moduleKind) {
+  return typeof moduleKind === 'string' && moduleKind.toLowerCase() === 'commonjs';
```
styfle, line 647: Do we need to handle the non-string case too?

src/index.js
```diff
@@ -258,6 +258,21 @@ function ncc (
...
+    target: 'esnext',
+    ...fullTsconfig.compilerOptions,
+    allowSyntheticDefaultImports: true,
+    noEmit: false,
+    outDir: '//'
+  };
+
+  if (esm && emitsCommonJsModule(tsCompilerOptions.module)) {
+    // ESM builds need TypeScript to preserve imports so webpack can resolve
+    // dependencies with "import" conditions instead of "require" conditions.
```
styfle, line 272: I'm still not sure this is the right solution.

Why should we change the tsconfig.json if the user explicitly set commonjs?

Does this even run when you try tsc + node? If it does then perhaps the code that detects the esm boolean is wrong above:

```
esm = entry.endsWith('.mjs') || !entry.endsWith('.cjs') && hasTypeModule(entry),
```

test/unit/ts-esm-import-condition/opt.json
```diff
@@ -0,0 +1,3 @@
+{
+  "esm": true
+}
```
styfle, line 3: Again, this is somewhat fictional. Why override the esm prop for a non-esm project?

Jaksenc:
Yeah you’re right, I think this still ended up testing a pretty weird shape.

I reran a small matrix after your comments and it made the issue clearer:

| case | main | this PR | `tsc + node` |
| --- | --- | --- | --- |
| forced `esm: true` + `module: commonjs`, no `type: module` | fails | passes | fails with `ERR_PACKAGE_PATH_NOT_EXPORTED` |
| real ESM TS: `type: module` + `module: esnext` | passes | passes | passes |
| `type: module` + `module: commonjs` | fails | passes | fails because CJS output runs in ESM scope |
| real ESM JS: `type: module`, normal import | passes | passes | n/a |

So yeah, the patch proves ncc *can* force this path to work, but I guess it doesn't really prove that this is a behavior ncc should support. The current version is basically overriding the user’s `module: commonjs` setting when `esm` is forced, and the cases it fixes don’t run cleanly with plain `tsc + node`.

Soo now im thinking that this probably shouldn’t land as-is? If you want ncc to support API-forced `esm: true` with CommonJS TS emit, I can reshape this around that as a explicit behavior. Otherwise I’m cool to close this one out since main already handles the normal ESM TS/JS shapes.

styfle:
> Soo now im thinking that this probably shouldn’t land as-is?

Agreed. We should probably remove the "forced esm" feature since ncc predates any formal esm configuration like `type:module` or `.mjs` so it needed its own config/option at the time. But now its just bloating the matrix of possibilities so we can remove in semver major.

PR #1334, opened 2026-06-14
fix: preserve runtime createRequire in ESM builds

Jaksenc:
## Problem

Since 0.38.2, an ESM build can replace a runtime
`createRequire(...)` binding with `undefined`, causing the later runtime
`require(...)` call to throw.

## Change

- Disable Webpack's `createRequire` parser handling in ncc so Node evaluates
  the call at runtime.
- Add a normal integration fixture with an `.mjs` entry that imports the
  TypeScript runtime payload and loads a runtime-selected JSON module.

This keeps the fix at ncc's Webpack configuration boundary and covers the
reported runtime behavior without changing dependency tracing.

## Validation

- `node scripts/build.js --no-cache`
- `node --expose-gc --max_old_space_size=4096 node_modules/jest/bin/jest.js test/integration.test.js --runInBand --testNamePattern create-require-runtime`
  - 1 passed
- `pnpm test -- test/unit.test.js`
  - 28 passed
- `node --expose-gc --max_old_space_size=4096 node_modules/jest/bin/jest.js --runInBand test/integration.test.js --testNamePattern='^(?!should execute "ncc run (?:axios\.js|canvas\.js|isomorphic-unfetch\.js|request-ts\.ts|request\.js|sharp\.js|socket\.io\.js)"$).*'`
  - 62 passed; 7 local environment-dependent cases skipped
- Negative control with Webpack's `createRequire` parsing restored reproduced
  `TypeError: input_require is not a function`.

## Risk

The parser option applies to all JavaScript handled by ncc. It intentionally
preserves `createRequire` for Node to evaluate at runtime, so modules selected
only at runtime must still be deployed with the bundle.

The cross-platform Node 22/24/26 matrix remains for CI. A local CLI watch test
failed with Watchpack `EMFILE` errors and reproduced identically on
`origin/main`, so it is not attributed to this change.

## Out of scope

- Changing how ncc traces or deploys runtime-selected modules.
- Dependency or lockfile updates.

## Issue linkage

Fixes #1312

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -385,7 +385,10 @@ function ncc (
         }
       ],
       parser: {
-        javascript: { importMeta: false },
+        javascript: {
+          createRequire: false,
+          importMeta: false
+        },
       },
     },
     plugins
--- a/test/integration/create-require-runtime.mjs
+++ b/test/integration/create-require-runtime.mjs
@@ -1,0 +1,1 @@
+import "./create-require-runtime/input.ts";
--- a/test/integration/create-require-runtime.mjs.stdout
+++ b/test/integration/create-require-runtime.mjs.stdout
@@ -1,0 +1,1 @@
+runtime-require-ok
--- a/test/integration/create-require-runtime/input.ts
+++ b/test/integration/create-require-runtime/input.ts
@@ -1,0 +1,15 @@
+import { createRequire } from "node:module";
+import { resolve } from "node:path";
+
+const require = createRequire(process.cwd());
+const runtimeModulePath = resolve(
+  process.cwd(),
+  "test/integration/create-require-runtime/runtime.json"
+);
+const { sentinel } = require(runtimeModulePath);
+
+if (sentinel !== "runtime-require-ok") {
+  throw new Error(`Unexpected runtime module value: ${sentinel}`);
+}
+
+console.log(sentinel);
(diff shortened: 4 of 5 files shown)


test/integration.test.js
```diff
@@ -83,6 +85,41 @@ for (const integrationTest of fs.readdirSync(__dirname + "/integration")) {
...
+    "integration/create-require-runtime.ts"
+  );
+  const tmpOut = fs.mkdtempSync(
+    path.join(os.tmpdir(), "ncc-create-require-runtime-")
+  );
+
+  try {
+    const { code, assets } = await ncc(input, {
+      cache: false,
+      esm: true,
```
styfle: Did you add the programatic api here instead of cli because you want to set `esm: true`?

If thats the only reason, its probably best to rename the `.ts` file to `.mts` or set `type:module` in `package.json` instead.

Jaksenc: Ohh yeah, that's exactly why I used the programmatic API there. I wanted to force `esm: true`, but you're right that it made the test more custom than it needed to be.

I moved it back into the normal integration path with a `.mjs` entry, and kept the actual `createRequire` bit in a tiny TS file that it imports.

I tried `.mts` too, but it still goes through the CJS path here, and typed `.mts` input doesn't seem to hit the TS loader. So `.mjs` felt like the smallest way to exercise the ESM path without adding `.mts` support in this PR.

Does that work better?

styfle (approved):
Great work, thanks!

PR #1337, opened 2026-07-29
fix: support TypeScript project references

### LINKED ISSUE #1212: Build fails with `composite: true` in `tsconfig.json`
I'm not sure if this is as designed or a bug. Any help is appreciated!

I have a npm workspace that uses TypeScript. All the packages have `composite: true` so they can be consumed by other packages using TypeScript `references`. 
When I try to use `ncc` to build one of the packages. the building fails. It succeeds if I set `composite` to `false`.

See the reproduction example in https://github.com/chubei-urus/ncc-demo.

JJordan0C:
## Summary
- allow ncc to bundle TypeScript composite projects that import referenced sibling sources
- keep composite build artifacts out of ncc's single-bundle compilation
- add a regression fixture matching the app/lib workspace reported in the issue
- run the regression through the standard unit golden-output runner in normal and coverage modes

## Root cause
When composite is enabled, TypeScript implicitly scopes rootDir to the directory containing the project tsconfig. ncc compiles the entry point and its imported sources as one webpack compilation, so a referenced sibling project was rejected with TS6059.

ncc already controls emit-related compiler options for its internal compilation. Disabling composite there preserves the single-bundle model while leaving the user's tsconfig unchanged.

## Tests
- pnpm build-test-binary
- pnpm build
- pnpm test (119 passed)
- pnpm test-coverage
- pnpm test test/unit.test.js --runInBand --testNamePattern=ts-project-references
- pnpm test-coverage test/unit.test.js --testNamePattern=ts-project-references

Fixes #1212

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -369,6 +369,9 @@ function ncc (
                 module: 'esnext',
                 target: 'esnext',
                 ...fullTsconfig.compilerOptions,
+                // ncc bundles the root project and its dependencies as one compilation.
+                // Disable composite's implicit rootDir so referenced sources are included.
+                composite: false,
                 allowSyntheticDefaultImports: true,
                 noEmit: false,
                 outDir: '//'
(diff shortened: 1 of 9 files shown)


styfle (approved):

styfle:
@JJordan0C Looks good - can you squash all the commits into one signed commit and force push?

